PR #254, opened 2015-02-02
Added set_pflying() function.

Nessiesson:


src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
```diff
@@ -4538,4 +4538,72 @@ public Version since() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{1, 2};
+		}
+
+		@Override
+		public String docs() {
+			return "void {[player], flight} Sets whether or not this player is flying. Requires player to have the ability to fly.";
```
LadyCailin: Maybe put how the player would get that ability (mention set_pflight)

src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
```diff
@@ -4538,4 +4538,72 @@ public Version since() {
 			return CHVersion.V3_3_1;
 		}
 	}
+	@api
```
LadyCailin: Add @seealso({set_pflight.class}), and then add @seealso({set_pflying.class}) to set_pflight. This will make them show up in each other's documentation.

PR #258, opened 2015-02-04
Properly support firework conversion and multiple effects.

jb-aero:


PseudoKnight:
This still breaks backwards compatibility, when it's entirely avoidable.

jb-aero:
If you got an email, that wasn't me pushing forward a fix, it just wouldn't let me switch branches without committing my local changes.

PseudoKnight:
It's a step closer, but it still fails in certain cases. You're forgetting about launch_firework, for one. I still strongly support using the firework: array(strength: 1, effects(array(), array(), array())) format for optimal compatibility.

jb-aero:
The way I've done it should accept that format, but it will tell them it is deprecated. Unless you mean writing to an array.

LadyCailin:
Just to clarify, you have more things you want to do with this PR, correct?

PseudoKnight:
It needs more things if he doesn't want to break scripts unnecessarily. And even if we were to ignore my disagreement with the array format (boo), he'd still need to support launch_firework since he wants to deprecate that format.

jb-aero:
Just going to close this until it's done.

PR #261, opened 2015-02-08
Added entity_interact event

PseudoKnight:
Like pressure_plate_activated, if you cancel this while a mob is standing on a pressure plate, it will naturally continue triggering the event every tick. Is this a concern or do we just let scripters handle it like other events? We can always add more prefilters later.


LadyCailin:
I think continual triggering is probably ok, especially if the event is being triggered by bukkit itself.

PR #263, opened 2015-02-12
Faster matching for procedures

PseudoKnight:
I discovered this simple little bottleneck when testing scripts with WarmRoast. 

When running the below test script, it consistently ran for 44-45 seconds before the optimization, and ran consistently for 33-34 seconds after the optimization. So over several runs it ran about 34% faster for this particular test, but given the location of the optimization we should see similar results for many scripts.

```
*:/test = >>>
	@start = time();
	for(@x = 0, @x < 300, @x++) {
		for(@y = 1, @y < 100, @y++) {
			for(@z = 0, @z < 300, @z++) {
				get_block_at(@x, @y, @z, 'world');
			}
		}
	}
	@stop = time();
	msg((@stop - @start).'ms');
<<<
```


LadyCailin:
Oh, very nice.

PR #265, opened 2015-02-12
Add support for reading UUIDs instead of a player name, and added more ...

jb-aero:
... string checking, fixes http://youtrack.sk89q.com/issue/CMDHELPER-2992.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCOfflinePlayer.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCOfflinePlayer.java
@@ -1,6 +1,8 @@
 
 package com.laytonsmith.abstraction;
 
+import java.util.UUID;
+
 /**
  *
  * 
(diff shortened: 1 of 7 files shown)


LadyCailin:
So, this is very good, though I have one concern with this, if getting the offline user may require a web hit, then the function calls need to be async, and take a closure callback. That way if the web hit is slow, it won't freeze the server. Thoughts?

jb-aero:
Funny enough, the methods I added are the ones where Bukkit/Spigot do NOT do a lookup. Anything that was using offline players before was already using the method that CAN do a lookup, but this should only happen if the server restarts after the player has been offline for 30 days and then one of the relevant functions is used on them. Specifically in Spikkit, getting a user by name should always be right, but may do a web lookup in the case mentioned above. Getting a player by uuid in Spikkit will never do a lookup currently, but if the player is not cached it may not return correct info.

VergilPrime:
LadyCailin plz.

PR #269, opened 2015-02-16
Added get_commands() and clear_commands()

Lildirt:


src/main/java/com/laytonsmith/core/functions/Commands.java
```diff
@@ -331,4 +335,120 @@ public Version since() {
...
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0};
+		}
+
+		@Override
+		public String docs() {
+			return "array {} Returns an array of command arrays in the format register_command expects."
+					+ " This does not include CommandHelper aliases, as they are not registered commands.";
```
LadyCailin: Use Implementation.GetServerType().getBranding() here instead of hardcoding "CommandHelper".

PR #272, opened 2015-02-28
Added server IP to get_server_info(12) which gives the current server IP

janwil:
Added server IP to get_server_info(12) which gives the current server IP


janwil:
Going to rerequest the pull request

LadyCailin:
Can you test this to make sure it works first?

PR #278, opened 2015-03-21
Implement version and mod-independent entity identification.

jb-aero:
Also fixed a long standing bug with get_server_info that probably nobody will notice.


jb-aero:
I've found a few more things that need to be changed.

jb-aero:
What this does:
* puuid() now accepts a boolean parameter which will remove the dashes from the uuid it returns.
* storm() with 3 arguments is now fixed, before it only assigned the world for 1 and 2 args.
* get_server_info() has been corrected to match the documentation. Indices 1 and 2 (Bukkit/API version and Craftbukkit/Mod version) were switched before. The internal method names have been renamed to clarify what each is.

And the main events:
* A dynamic entitytypes list has been implemented that populates itself at server start. Our names for vanilla mobs are matched to mapped to the vanilla Bukkit names, and any modded entities added by Forge are added mapped to their own name, marking themselves as the Unknown type from Bukkit. It will not create entries for entities that do not exist on the server's Minecraft version. BukkitConvertor's BukkitGetCorrectEntity() method now uses the dynamic list to reflectively obtain the appropriate class. If the mob has a specialized class created by us, that will be used, otherwise it will try to put it into a general class using the old method. The only maintenance this should need is adding new vanilla types to that list, and if we add a specialized class that doesn't follow the conventional naming scheme, it will need an exception in the reflection method. All the information for this system is cached, it is generated on startup, and the only processing at runtime is retrieving the information from memory. The result is a slight increase in startup time, but the latest version of CH will be able to operate on entities without giving the server entities the server doesn't know about, and without the server giving CH entities CH doesn't know about, with theoretically no runtime performance hit (tested on a 1.7.10 Forge server with 8-10 players).

* A dynamic event registrar has been added as well. Basically it won't register events that don't exist yet on the server's Minecraft version. This looks complicated, but the only thing it does that it didn't before is compare the server version to the version each event was added.

With these changes, I've encountered NO errors running the master version of CH on Cauldron for 1.7.10 or Spigot for 1.8. Testing MCPC+ for 1.6.4 will come later, but the goal is to no longer need backports.

src/main/java/com/laytonsmith/abstraction/enums/MCVersion.java
```diff
@@ -0,0 +1,48 @@
+package com.laytonsmith.abstraction.enums;
+
+/**
+ * Created by jb_aero on 3/17/2015.
+ * <p/>
+ * This may require unforeseen changes later to support fossil versions, but odds are support will not be requested.
+ */
+public enum MCVersion {
```
LadyCailin: implements Version. That will make it a little more useable in other places

src/main/java/com/laytonsmith/annotations/EventIdentifier.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Created by jb_aero on 3/22/2015.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface EventIdentifier {
+
+	String className();
+
+	String priority();
```
LadyCailin: Why is this a string, and not an enum?

jb-aero: Because Sponge uses a completely different names and that seemed an acceptable way of matching the enum of either API, given only those of us adding new events will use it, and only when we add events that don't exist on other versions. The end user will never have to match something to that, and the value is only used on plugin load.

PR #283, opened 2015-04-09
Fixed error in core caused by ptake_item.

Pieter12345:
ptake_item(@player, null, @amount) returned an error in core
(NullPointerException in the match() method). It will now throw a
ConfigRuntimeException with type IllegalArgumentException instead.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
+++ b/src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
@@ -792,9 +792,6 @@ public class InventoryManagement {
                 p = Static.GetPlayer(args[0], t);
                 is = Static.ParseItemNotation(this.getName(), args[1].val(), Static.getInt32(args[2], t), t);
             }
-            if(is == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
```diff
@@ -821,7 +818,13 @@ public Construct exec(Target t, Environment environment, Construct... args) thro
         }
 
         private boolean match(MCItemStack is, MCItemStack iis){
-            return (is.getTypeId() == iis.getTypeId() && is.getData().getData() == iis.getData().getData());
+            if(is.getData() == null && iis.getData() == null) {
+                return is.getTypeId() == iis.getTypeId();
```
jb-aero: Since you're changing this now outdated line, I'd HIGHLY advise switching to getType(), as I'm going to have to change it to that anyway because handling items by numeric IDs is no longer guaranteed, and thus there is not even an equivalent function in Sponge.

PR #286, opened 2015-04-14
Added note about @arguments being unavailable for use in this instance.

VergilPrime:


PseudoKnight:
As I mentioned in the Issue, another possibility is to change the order so that a coder defined @arguments takes precedence when passed to a closure. It should be simple, if I'm not mistaken. I'll leave the choice up to better minds.

Could be as easy as moving this above line 136:
https://github.com/sk89q/CommandHelper/blob/master/src/main/java/com/laytonsmith/core/constructs/CClosure.java#L149-L155

LadyCailin:
Actually, it might be nice to issue a warning, informing the users when they override @arguments in any case. Any thoughts on that?

PseudoKnight:
If they override it, then they're probably not using the original. (I don't think I've ever seen anyone use it; I know I haven't) Consider it like a default value. However, as it is, when @arguments overrides explicit script, that's a problem.

LadyCailin:
Well, I agree with that, the current behavior is certainly not desired. I think though, that the desired behavior here would be to allow it, but with a warning. Ideally, they wouldn't override the value, though it is clear what they are intending.

LadyCailin:
I have fixed this in a more generic way. It is now allowed to override @arguments, but a warning is issued. I am going to close this PR, as it is no longer necessary.

PR #287, opened 2015-04-14
Fixed array_get's deep cloning.

Pieter12345:
- array_get(@array) or @array[] only cloned 2 arrays deep. It will now recursively clone the whole array.
- Added array_deep_clone(@array).
- Added array_shallow_clone(@array).

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
@@ -146,6 +146,8 @@ public class ArrayHandling {
 					// Deep clone the array if the "index" is the initial one.
 					if (((CSlice) index).getStart() == 0 && ((CSlice) index).getFinish() == -1) {
 						return ca.deepClone(t);
+					} else if(ca.inAssociativeMode()) {
+						throw new ConfigRuntimeException("Array slices are not allowed with an associative array", ExceptionType.CastException, t);
 					}
 					
 					//It's a range
--- a/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
@@ -2248,12 +2248,21 @@ public class ArrayHandling {
 			return CHVersion.V3_3_1;
 		}
 
-//		@Override
-//		public ExampleScript[] examples() throws ConfigCompileException {
-//			return new ExampleScript[]{
-//				new ExampleScript("", "")
-//			};
-//		}
+		@Override
+		public ExampleScript[] examples() throws ConfigCompileException {
+			return new ExampleScript[]{
+				new ExampleScript("Demonstrates that the array is cloned.",
+						"@array = array(1, 2, 3, 4)\n" +
+						"@deepClone = array_deep_clone(@array)\n" +
+						"@deepClone[1] = 'newValue'\n" +
+						"msg(@array)\nmsg(@deepClone)"),
+				new ExampleScript("Demonstrated that arrays within the array are also cloned by a deep clone.",
+						"@array = array(array('value'))\n" +
+						"@deepClone = array_deep_clone(@array)\n" +
+						"@deepClone[0][0] = 'newValue'\n" +
+						"msg(@array)\nmsg(@deepClone)")
+			};
+		}
 
 	}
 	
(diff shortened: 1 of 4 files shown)


src/test/java/com/laytonsmith/core/functions/ArrayHandlingTest.java
```diff
@@ -306,4 +305,20 @@ public void testStringSlice() throws Exception {
 	@Test public void testArrayUnique4() throws Exception {
 		assertEquals("{1, 1}", SRun("array_unique(array(1, '1', 1), true)", fakePlayer));
 	}
+	
+	@Test public void testArrayGetClone() throws Exception {
+		Run("@a = array(array(array('value'))); @b = @a[]; @b[0][0][0] = 'changedValue'; msg(@a[0][0][0]);", fakePlayer);
+		verify(fakePlayer).sendMessage("value");
```
LadyCailin: This should also msg @b[0][0][0] to insure that it is 'changedValue. Add the msg(@b...) to the Run, then add another verify(...) under line 311.

src/test/java/com/laytonsmith/core/functions/ArrayHandlingTest.java
```diff
@@ -306,4 +305,20 @@ public void testStringSlice() throws Exception {
...
+	}
+	
+	@Test public void testArrayGetCloneRefCouples() throws Exception {
+		Run("@a = array('Meow'); @b = array(@a, @a, array(@a)); @c = @b[]; msg((ref_equals(@b[0], @b[1]) && ref_equals(@b[0], @b[2][0])));", fakePlayer);
```
LadyCailin: @c is not used here, I think you meant for it to be.

src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
```diff
@@ -144,21 +142,14 @@ public Construct exec(Target t, Environment env, Construct... args) throws Cance
...
-									na.set(key, ca.get(key, t).clone(), t);
-								} catch (CloneNotSupportedException ex) {
-									na.set(key, ca.get(key, t), t);
-								}
-							}
-							return na;
-						}
+						
+					// Deep clone the array if the "index" is the initial one.
+					if (((CSlice) index).getStart() == 0 && ((CSlice) index).getFinish() == -1) {
```
LadyCailin: Remove this case. I will make it so that @a[] calls array_deep_clone instead, so this will only be triggered when the user is doing a slice, which will remain a shallow copy.

src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
```diff
@@ -2207,56 +2198,125 @@ public Version since() {
...
+
+		@Override
+		public Version since() {
+			return CHVersion.V3_3_1;
+		}
+
 //		@Override
-//		public Version since() {
-//			return CHVersion.V3_3_1;
+//		public ExampleScript[] examples() throws ConfigCompileException {
```
LadyCailin: Can you add an example?

src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
```diff
@@ -2207,56 +2198,125 @@ public Version since() {
...
+			return "array {array} Performs a shallow clone on an array (as opposed to a deep clone). See the examples for more info.";
+		}
+
+		@Override
+		public Version since() {
+			return CHVersion.V3_3_1;
+		}
+
 //		@Override
 //		public ExampleScript[] examples() throws ConfigCompileException {
```
LadyCailin: An example here too

src/test/java/com/laytonsmith/core/functions/ArrayHandlingTest.java
```diff
@@ -306,4 +305,35 @@ public void testStringSlice() throws Exception {
 	@Test public void testArrayUnique4() throws Exception {
 		assertEquals("{1, 1}", SRun("array_unique(array(1, '1', 1), true)", fakePlayer));
 	}
+	
+	@Test public void testArrayGetClone() throws Exception { // This is expected to be a deep clone.
```
LadyCailin: Once you make the change above, this will fail. That's ok, I'll fix it.

LadyCailin:
Oh, one thing I forgot to have you add is a @seealso annotation. That links the functions in the documentation. Not super important though.

PR #290, opened 2015-04-17
Add automatic code style formatting

jacobwgillespie:
Adds the maven java formatter plugin and integrates it automatically with the build system.  Formatting occurs on all builds or on-demand with `mvn java-formatter:format`.

Cheers!


LadyCailin:
I'm going to reject this PR, because this formatter uses the Eclipse style, which I do not like.

PR #291, opened 2015-04-18
Add uptime and gc stats to get_server_info

jacobwgillespie:
This is a partial implementation of http://youtrack.sk89q.com/issue/CMDHELPER-2641

TPS still isn't there, and it looks like it takes a loop to measure it (Essentials implementation here: https://github.com/essentials/Essentials/blob/a2c43d822c66e617a84df9a8f074b9c3a3e32fae/Essentials/src/com/earth2me/essentials/EssentialsTimer.java).  Not sure if we want to do that or not.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/Minecraft.java
+++ b/src/main/java/com/laytonsmith/core/functions/Minecraft.java
@@ -812,9 +812,9 @@ public class Minecraft {
 					+ "<li>11 - Server port; Get the game port that the server runs on</li>"
 					+ "<li>12 - Server IP; Get the IP that the server runs on</li>"
 					+ "<li>13 - Uptime; The number of milliseconds the server has been running</li>"
-					+ "<li>14 - gcmax; The maximum amount of memory that the Java virtual machine will attempt to use</li>"
-					+ "<li>15 - gctotal; The total amount of memory in the Java virtual machine</li>"
-					+ "<li>16 - gcfree; The amount of free memory in the Java Virtual Machine</li></ul>";
+					+ "<li>14 - gcmax; The maximum amount of memory that the Java virtual machine will attempt to use, in bytes</li>"
+					+ "<li>15 - gctotal; The total amount of memory in the Java virtual machine, in bytes</li>"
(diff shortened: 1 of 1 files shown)


jb-aero:
I'd say leave dividing the memory to the user, as doing it beforehand gives them less info.

jacobwgillespie:
Ah, good idea - removed it and added a note to the docs on the returned units.

src/main/java/com/laytonsmith/core/functions/Minecraft.java
```diff
@@ -922,6 +929,23 @@ public Construct exec(Target t, Environment env, Construct... args) throws Cance
 				//Server Ip
 				retVals.add(new CString(server.getIp(), t));
 			}
+			if (index == 13 || index == -1) {
+				//Uptime
+				double uptime = (double)(new GregorianCalendar().getTimeInMillis() - ManagementFactory.getRuntimeMXBean().getStartTime());
```
LadyCailin: Is there a reason System.currentTimeMillis() won't work here instead of instantiating a new object?

jb-aero: He was using it here because Essentials uses it, but Essentials does so to format it, which we already have a function for.

jacobwgillespie: Ah, cool - better to have fewer instantiations :)

src/main/java/com/laytonsmith/core/functions/Minecraft.java
```diff
@@ -922,6 +929,23 @@ public Construct exec(Target t, Environment env, Construct... args) throws Cance
 				//Server Ip
 				retVals.add(new CString(server.getIp(), t));
 			}
+			if (index == 13 || index == -1) {
+				//Uptime
+				double uptime = (double)(System.currentTimeMillis() - ManagementFactory.getRuntimeMXBean().getStartTime());
+				retVals.add(new CDouble(uptime, t));
```
jb-aero: I didn't notice this before, but why is this a double and CDouble instead of long and CInt?

PR #293, opened 2015-04-20
Allow disabling of potion particles

jacobwgillespie:
This allows an optional sixth parameter to `set_mob_effect()` and `set_peffect()` to configure whether particles are visible at all.  The `particles` value is also returned from `get_peffects()` and `get_mob_effects()` as a boolean.

Since I'm not 100% sure on backwards compatibility of the parameter inside Bukkit, the 5-parameter constructor for `PotionEffect` is only used if specifically requested.

Documentation has been added for all changes.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
@@ -75,11 +75,21 @@ public interface MCLivingEntity extends MCEntity, MCProjectileSource {
 		private int strength;
 		private int secondsRemaining;
 		private boolean ambient;
+		private boolean particles;
 		public MCEffect(int potionID, int strength, int secondsRemaining, boolean ambient){
 			this.potionID = potionID;
 			this.strength = strength;
 			this.secondsRemaining = secondsRemaining;
 			this.ambient = ambient;
+			this.particles = true;
+		}
+
+		public MCEffect(int potionID, int strength, int secondsRemaining, boolean ambient, boolean particles){
+			this.potionID = potionID;
+			this.strength = strength;
+			this.secondsRemaining = secondsRemaining;
+			this.ambient = ambient;
+			this.particles = particles;
 		}
 
 		public int getPotionID() {
@@ -97,6 +107,8 @@ public interface MCLivingEntity extends MCEntity, MCProjectileSource {
 		public boolean isAmbient() {
 			return ambient;
 		}
+
+		public boolean hasParticles() { return particles; }
 
 	}
 }
--- a/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
@@ -13,6 +13,7 @@ import java.util.List;
 public interface MCLivingEntity extends MCEntity, MCProjectileSource {
 
 	public void addEffect(int potionID, int strength, int seconds, boolean ambient, Target t);
+	public void addEffect(int potionID, int strength, int seconds, boolean ambient, boolean particles, Target t);
 	public boolean removeEffect(int potionID);
 	/**
 	 * Returns the maximum effect id, inclusive.
--- a/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
@@ -108,7 +108,9 @@ public interface MCLivingEntity extends MCEntity, MCProjectileSource {
 			return ambient;
 		}
 
-		public boolean hasParticles() { return particles; }
+		public boolean hasParticles() {
+			return particles;
+		}
 
 	}
 }
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
@@ -221,6 +221,16 @@ public class BukkitMCLivingEntity extends BukkitMCEntityProjectileSource impleme
 		}
 	}
 
+	/**
+	 * This calls the 5-param constructor for PotionEffect to disable particles entirely
+	 * @param potionID - ID of the potion
+	 * @param strength - potion strength
+	 * @param seconds - duration of the potion in seconds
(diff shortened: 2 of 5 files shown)


src/main/java/com/laytonsmith/abstraction/MCLivingEntity.java
```diff
@@ -97,5 +108,7 @@ public boolean isAmbient() {
 			return ambient;
 		}
 
+		public boolean hasParticles() { return particles; }
```
LadyCailin: Put this on 3 lines, not 1

src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
```diff
@@ -216,31 +216,22 @@ public void addEffect(int potionID, int strength, int seconds, boolean ambient,
...
+
+	public void addEffect(int potionID, int strength, int seconds, boolean ambient, boolean particles, Target t) {
+		PotionEffect pe = new PotionEffect(PotionEffectType.getById(potionID), (int)Static.msToTicks(seconds * 1000),
+				strength, ambient, particles);
+		try{
+			if(le != null){
+				le.addPotionEffect(pe, true);
+			}
+		} catch(NullPointerException e){
+			Logger.getLogger(BukkitMCLivingEntity.class.getName()).log(Level.SEVERE,
```
LadyCailin: Can you put a comment here explaining this?

src/main/java/com/laytonsmith/core/ObjectGenerator.java
```diff
@@ -835,6 +835,7 @@ public CArray potions(List<MCLivingEntity.MCEffect> effectList, Target t) {
 			effect.set("strength", new CInt(eff.getStrength(), t), t);
 			effect.set("seconds", new CInt(eff.getSecondsRemaining(), t), t);
 			effect.set("ambient", CBoolean.get(eff.isAmbient()), t);
+			effect.set("particles", CBoolean.get(eff.hasParticles()), t);
```
jb-aero: This also needs to be added to the CArray -> effectList converter.

src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
jacobwgillespie: Yep, added.

src/main/java/com/laytonsmith/core/ObjectGenerator.java
jacobwgillespie: Done - thanks!

PseudoKnight:
Just to confirm, this was added in 1.8.

src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
LadyCailin: Um, not sure you actually did add it.

src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCLivingEntity.java
```diff
@@ -216,31 +216,32 @@ public void addEffect(int potionID, int strength, int seconds, boolean ambient,
...
+	 * @param strength - potion strength
+	 * @param seconds - duration of the potion in seconds
+	 * @param ambient - make particles less noticable
+	 * @param particles - enable or disable particles entirely
+	 * @param t - target
+	 */
+	@Override
+	public void addEffect(int potionID, int strength, int seconds, boolean ambient, boolean particles, Target t) {
+		PotionEffect pe = new PotionEffect(PotionEffectType.getById(potionID), (int)Static.msToTicks(seconds * 1000),
+				strength, ambient, particles);
```
jb-aero: To support versions prior to 1.8, change this to:
```
PotionEffect pe;
if (Static.getServer().getMinecraftVersion().lt(MCVersion.MC1_8)) {
	pe = new PotionEffect(PotionEffectType.getById(potionID), (int)Static.msToTicks(seconds * 1000),
			strength, ambient);
} else {
	pe = new PotionEffect(PotionEffectType.getById(potionID), (int)Static.msToTicks(seconds * 1000),
			strength, ambient, particles);
}
```
Then there doesn't need to be two different methods, and it won't error on 1.7 if someone tries to use the particles option. There needs to be a similar change on the getEffects method, because a 1.7 server will not like pe.hasParticles(). If the version is less than 1.8, just construct the MCEffect with "true" instead.

src/main/java/com/laytonsmith/core/ObjectGenerator.java
```diff
@@ -861,7 +863,14 @@ public CArray potions(List<MCLivingEntity.MCEffect> effectList, Target t) {
...
 				}
-				ret.add(new MCLivingEntity.MCEffect(potionID, strength, seconds, ambient));
+				if (effect.containsKey("particles")) {
+					particles = Static.getBoolean(effect.get("particles", t));
+				}
+				if (particles) {
+					ret.add(new MCLivingEntity.MCEffect(potionID, strength, seconds, ambient));
+				} else {
+					ret.add(new MCLivingEntity.MCEffect(potionID, strength, seconds, ambient, particles));
+				}
```
jb-aero: This if statement is unnecessary.

src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
```diff
@@ -2090,17 +2092,21 @@ public Construct exec(Target t, Environment env, Construct... args) throws Confi
 			int strength = Static.getInt32(args[2], t);
 			int seconds = 30;
 			boolean ambient = false;
+			boolean particles = true;
 			if (args.length >= 4) {
 				seconds = Static.getInt32(args[3], t);
 			}
 			if (args.length == 5) {
```
LadyCailin: This needs to be changed, as you still want to read in args[4] if the length is either 5 OR 6. Change it to args.length >= 5. Also, use args.length >= 6 below, just for good measure.

jacobwgillespie: Good catch - fixed.

PR #294, opened 2015-04-20
AliasCore GlobalEnv update (bugfix in set_uncaught_exception_handler())

Pieter12345:
AliasCore's alias() method created a new GlobalEnv every time it was called. It will now keep using the same instance of GlobalEnv after the first alias() call.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ true
 /preferences.*
 .DS_Store
 /test-backend/
+persistence.json
--- a/src/main/java/com/laytonsmith/PureUtilities/Common/MutableObject.java
+++ b/src/main/java/com/laytonsmith/PureUtilities/Common/MutableObject.java
@@ -37,7 +37,7 @@ public final class MutableObject<T> {
 	 * @return
 	 */
 	public T getObject(){
-		return obj;
+		return this.obj;
 	}
 
 	@Override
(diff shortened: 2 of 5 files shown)


LadyCailin:
I'm going to pull this locally and play around with it for a bit, before I merge it, so just letting you know this PR will take longer than most.

PseudoKnight:
This can't work. Each alias needs its own environment for iVariableList, procs, etc. Shouldn't uncaughtExceptionHandler be shared among all environments, like it says? What's the bug?

Pieter12345:
@PseudoKnight This is a very old PR that was created because set_uncaught_exception_handler() in interpreter would not work for script errors as the interpreter environment was a different environment than the script's environment. I believe LadyCailin said that interpreter should be a sandbox which should have its own environment months ago, I think I reverted that part but I'm not sure about that. It still left the problem that environments were not shared, and the uncaught exception handler could not be set globally. This PR solved that but I have no idea what LadyCailin wanted with that.

@LadyCailin What are your ideas on this PR? Is it just forgotten or is it just unknown what side effects this would have or should the uncaught exception handler just be set per script instead of globally?

PseudoKnight:
http://dev.enginehub.org/youtrack/issue/CMDHELPER-3162

PR #295, opened 2015-04-21
Expose item_flags meta value

jacobwgillespie:
This exposes item flags as a meta value in item arrays.

This closes http://youtrack.sk89q.com/issue/CMDHELPER-3036

https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/ItemFlag.html

:cake:

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCItemMeta.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCItemMeta.java
@@ -4,6 +4,8 @@
  */
 package com.laytonsmith.abstraction;
 
+import com.laytonsmith.abstraction.enums.MCItemFlag;
+
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


PseudoKnight:
Haven't looked at all the code, I just wanted to comment that "item_flags" is redundant. Perhaps just "flags" instead? It's not clear if Spigot named the "item flags" because they plan to store other types of flags in there, but right now they're just the new 1.8 HideFlags.

PseudoKnight:
Though their description says "A ItemFlag can hide some Attributes from ItemStacks", so it sounds like it'll just be HideFlags.

jacobwgillespie:
Updated - I left the variables as item_flags since that's closer to internals, but it's exposed as "flags".

PseudoKnight:
If the ENUM didn't have "HIDE_" in it, I might have recommended "'hide': array('ATTRIBUTES', 'ENCHANTS')". Sometimes it's curious what different developers choose for naming.

LadyCailin:
Can you add the information about this new parameter to the documentation for the appropriate functions that are affected by this?

PseudoKnight:
Should this be null if there's no flags? These will be rare and it will be even more rare to check for them, rather than just setting them. So it makes sense to use null instead of an empty CArray, imo, unlike enchants.

LadyCailin:
An empty array is fine, since array() == false, and also triggers the else condition on a foreach else.

PseudoKnight:
True, but null is lighter than a CArray, which seems worthwhile given the use cases?

LadyCailin:
Well, I guess so. Though we should make special note that it's different than enchantments, since it won't be consistent.

PseudoKnight:
It's consistent with lore, though.

LadyCailin:
Lol. Ok, well, whatever then.

PseudoKnight:
Heh. Well, if it was completely up to me I wouldn't have even included the indexes if it didn't have the data. I ended up creating a _minify_inv() procedure for storage and display, instead of the increasingly large item arrays we have now. I mean, enchants are in there twice. lol

jacobwgillespie:
Okay, added docs - this is ready for review.

LadyCailin:
Closing this, as I just pulled Pieter's, which does the same thing.

PR #299, opened 2015-04-23
Multiple updates, fixes, and optimizations to Vector3D, Point3D, and refactors to all classes using them.

theoctopod:
Basically, the entire point of this PR is to set up for the point/vector functions I plan to make.

- `Point3D`, previously located in the `PureUtilities.Geometry` class has been moved up into the `PureUtilities` package, with all references updated.
  - `Point3D` is now immutable. All methods are fixed to return new objects instead of modifying the current object's variables.
  - Deprecated the zero constructor. Use `Point3D.ZERO` instead.
  - Added a copy constructor and 2D constructor.
  - Added `translate()` which will translate the current point along the values of another point.
  - Split up `distance()` into `distance()` and `distanceSquared()`.
  - All methods and constructors now have documentation.
- `MVector3D`, previously located in the `abstraction` package is now called `Vector3D` located in the `PureUtilities` package, with all references updated.
  - `Vector3D` is now immutable. All methods will return new objects instead of modifying the current object's variables.
  - `Vector3D` is now subclass of `Point3D`. All point-related methods have been moved to `Point3D`.
  - Deprecated the zero constructor. Use `Vector3D.ZERO` instead.
  - Added a copy constructor and 2D constructor.
  - Added `divide()`, which seemed fitting seeing as there's an `add()`, `substract()`, and `multiply()`
  - Added `dot()` and `cross()`, which will return the dot and cross products of the vector.
  - All methods and constructors now have documentation.
- `PureUtilities.Geometry` has been deleted.
- Added/Rewrote the `vector()` conversion methods. Vectors can be formatted as `array(1, 2, 3)` or `array(x: 1, y: 2, z: 3)` in MethodScript according to these methods.
  - `CArray vector(Vector3D, Target t)` has been deprecated as there's no need for a Target in this conversion. (the location conversion method didn't have it) `CArray vector(Vector3D)` is recommended instead.
  - added `Vector3D vector(CArray, Target t)` conversion method.


theoctopod:
also, there's actually no Vector3D.ZERO yet. whoops.

theoctopod:
how did the IDE miss the compiler error that would've made?

src/main/java/com/laytonsmith/PureUtilities/Point3D.java
```diff
@@ -0,0 +1,127 @@
+package com.laytonsmith.PureUtilities;
```
LadyCailin: Can you fix the spacing in this file?

src/main/java/com/laytonsmith/PureUtilities/Point3D.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Represents a point in 3D space.
+ *
+ */
+public class Point3D {
+
+	/**
+	 * A Point3D located at [0, 0, 0] (zero)
+	 */
+	public static Point3D ZERO = new Point3D();
```
LadyCailin: public static final would be better here.

src/main/java/com/laytonsmith/PureUtilities/Vector3D.java
```diff
@@ -0,0 +1,162 @@
+package com.laytonsmith.PureUtilities;
+
+/**
+ * Represents both a point in 3D space and a vector representing a direction and magnitude.
+ * 
+ */
+public class Vector3D extends Point3D {
+
+	public static Vector3D ZERO = new Vector3D();
```
LadyCailin: public static final

src/main/java/com/laytonsmith/PureUtilities/Point3D.java
```diff
@@ -0,0 +1,127 @@
...
+
+    protected double x;
```
LadyCailin: Since these are immutable, you should make these properties final, so they can't be changed.

PR #305, opened 2015-04-30
Removed deprecated constructors + minor changes.

Pieter12345:
- Removed (the only) use of the deprecated Vector3D and Point3D and their constructors.
- Added javadoc to Vector3D.ZERO.
- Redirected all constructors to class..(x, y, z).

### DIFF
--- a/src/main/java/com/laytonsmith/PureUtilities/Point3D.java
+++ b/src/main/java/com/laytonsmith/PureUtilities/Point3D.java
@@ -9,7 +9,7 @@ public class Point3D
 	/**
 	 * A Point3D located at [0, 0, 0] (zero)
 	 */
-	public static final Point3D ZERO = new Point3D();
+	public static final Point3D ZERO = new Point3D(0, 0, 0);
 
 	protected final double x;
 	protected final double y;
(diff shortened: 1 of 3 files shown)


src/main/java/com/laytonsmith/PureUtilities/Point3D.java
```diff
@@ -9,7 +9,7 @@
 	/**
 	 * A Point3D located at [0, 0, 0] (zero)
 	 */
-	public static final Point3D ZERO = new Point3D();
+	public static final Point3D ZERO = new Point3D(0, 0, 0);
```
theoctopod: well, the reason I wanted to privatize Point3D() was so the only place it would be run was from this line. 
if you end up using Point3D(0, 0, 0), what's the point? nothing else in this class uses that constructor.

Pieter12345: Ah, we had a misunderstanding then. I didn't quite understand why I should not remove it. I'll remove them then.

src/main/java/com/laytonsmith/PureUtilities/Point3D.java
```diff
@@ -27,25 +27,10 @@ public Point3D(Point3D other) {
...
-	}
-
-	/**
 	 * Initializes the X and Y values. Z is initialized to 0.
 	 */
 	public Point3D(double x, double y) {
-		this.x = x;
-		this.y = y;
-		this.z = 0;
+		this(x, y, 0);
```
theoctopod: could you please revert this to how it used to be?
http://stackoverflow.com/questions/30001991/performance-should-i-avoid-constructor-delegation
just a tiny tiny optimization.

Pieter12345: The difference is very minor, but I changed it back for you :)

PR #306, opened 2015-05-13
Fixes CMDHELPER-3060

PseudoKnight:
Tested on 1.7.2 and 1.8.3 just to make sure it's all working this time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,8 +173,8 @@
 
 		<!-- Bukkit -->
 		<dependency>
-			<groupId>org.bukkit</groupId>
-			<artifactId>bukkit</artifactId>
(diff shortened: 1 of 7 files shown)


PseudoKnight:
Oh, I didn't use invokeMethod() here, but is that really needed? It seemed to test fine in 1.7.2.

PseudoKnight:
It seems like it only ever could be an issue if they removed the deprecated methods. But they're only deprecated in Spigot right now.

PhanaticD:
pls yes

PseudoKnight:
Alternatively we could continue to build against Bukkit and I could use invokeMethod on the Spigot-only methods as well.

PR #307, opened 2015-05-17
Remove deprecated and slow EventUtils.TriggerExternal() from listeners

PseudoKnight:
I've been doing some profiling when I noticed that TriggerExternal(), though seldom used and no longer implemented in some listeners, creates a noticeable overhead. While the operation times vary, it's easy to pick out a pattern where listeners that don't use TriggerExternal() have a base operation time that is ~80-90% less than those that do.

Example Test 1: 
unbound events w/ TriggerExternal() = 0.10-0.14ms
unbound events w/o TriggerExternal() = 0.01-0.02ms

Example Test 2:
unbound events w/ TriggerExternal() = 0.18-0.24ms
unbound events w/o TriggerExternal() = 0.02-0.04ms

I also verified this conclusion by timing each part of the events stack, where it consistently showed TriggerExternal() taking hundreds of microseconds on my test server. (above examples on my faster production server)

TriggerExternal() has been deprecated for some time now, so I propose we update the affected extensions and remove this method and all its uses from CommandHelper. I looked at all the extensions on the build site and found that the functionality is only used by CHAdvanced and CHVirtualChests over 3 events. I recommend this not be pulled until those have been modified.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCCommand.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCCommand.java
@@ -232,7 +232,6 @@ public class BukkitMCCommand implements MCCommand {
 			}
 		}
 		BukkitMCCommandTabCompleteEvent event = new BukkitMCCommandTabCompleteEvent(sender, cmd, alias, args);
-		EventUtils.TriggerExternal(event);
 		EventUtils.TriggerListener(Driver.TAB_COMPLETE, "tab_complete_command", event);
 		return event.getCompletions();
 	}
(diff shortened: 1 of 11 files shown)


EntityReborn:
I'm ok with this, with one request.
This will break CHVC. Pending this request being accepted by LC, I'd appreciate a patch to CHVirtualChests. I don't have time or things set up to do this myself, and it's not quite fair that this change affect the many users of CHVC.

jb-aero:
Off the top of my head, I expect this would also break SKCompat's ability to let WorldGuard block aliases. It's there so that extensions can have a say in standard events.

PseudoKnight:
EntityReborn, I was already planning on it as per my last two sentences. I just needed to get feedback if I was missing anything. I considered just leaving the inv open and inv close events intact for a while longer for a smoother transition. But extensions have had to be updated frequently in recent months anyway.

Creating a new wrapper object is much much faster than the using TriggerExternal(). It seems it only does any heavy lifting the first time, and then each subsequent event it takes something like several microseconds. So it'd still make extensions that used TriggerExternal() take 70-80% shorter time.

SKCompat is sort of a special case (probably why I forgot about it). It doesn't need a CH event and it needs to be cancelled before it fires an alias. Can we add a traditional event here?

PseudoKnight:
I'm thinking I should just re-submit a commit that keeps TriggerExternal() for now and removes uses that extensions have not used. (so, inv open/close, player command, and server command) Then we can address the remaining uses in time. (i have a couple thoughts for player command)

PR #313, opened 2015-06-11
Code cleanup + minor changes.

Pieter12345:
- Removed unused imports and initializations.
- Replaced ){ like notation by ) { like notation.
- Changed some assigns that overwrite eachother to if/else statements.


Pieter12345:
This PR is way too old now. Merging would take more effort than re-doing it.

PR #315, opened 2015-07-08
Added ItemFlag support to set/get_itemmeta.

Pieter12345:
set_itemmeta() and get_itemmeta() now accept array('itemflags':
array('FLAG1', 'FLAG2', 'etc')) with all ItemFlag values currently
supported by bukkit.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCItemMeta.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCItemMeta.java
@@ -4,8 +4,10 @@
  */
 package com.laytonsmith.abstraction;
 
+import com.laytonsmith.abstraction.enums.MCItemFlag;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  *
@@ -77,6 +79,18 @@ public interface MCItemMeta extends AbstractionObject {
      * @return whether the enchantment was added successfully
      */
     boolean addEnchant(MCEnchantment ench, int level, boolean ignoreLevelRestriction);
+	
+	/**
+	 * Set itemflags which should be ignored when rendering a MCItemStack in the Client.
+	 * @param flags The flags to ignore.
+	 */
+	void addItemFlags(MCItemFlag... flags);
+	
+	/**
+	 * Get current set itemFlags.
+	 * @return A set of all itemFlags set
+	 */
+	public Set<MCItemFlag> getItemFlags();
     
     /**
      * Removes a given enchantment from this meta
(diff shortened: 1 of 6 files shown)


PseudoKnight:
Pieter, you should take a look at #295

Pieter12345:
@PseudoKnight Oh sigh, double work, this is tested and does work fine though.

PseudoKnight:
I looked it over. I like this PR a little better. I do have some thoughts though. (not sure how academic this will be)

- As mentioned in the other PR "itemflags" is a redundant key in an item array. "flags" or "hideflags" would probably be more appropriate. "hide" would seem good too, if the ENUMs didn't start with HIDE_.

- In this PR the index doesn't exist if there's no hide flags. The minimalist in me likes this, but it's inconsistent with how item meta is handled elsewhere. The rule is always include the index if the item can handle the meta (eg. only books have the pages index). Now whether it should be null or an empty array is up to debate. Enchants use an empty array, but everything else uses null, so I probably recommend null.

Pieter12345:
@PseudoKnight
I agree with renaming "itemflags" to "flags". I'd prefer to keep the enum the same, as it's 1:1 with bukkit's ItemFlag now.

About the indices:
- "firework" accepts an array and defaults to null.
- "display" accepts a string and defaults to null.
- "potions" doesn't appear when the item is not a potion.
- "enchants" accepts an array and default to array().

I believe we just had all options there. To implement this, I could do several things:

- One option could be to change the "enchants" to default to null and "fireworks" to be only included for actual firework meta and add "flags" with a null default.

- Another option (I like this best) could be to not include any index that is empty. This allows users to check if the item has a metadata using array_index_exists instead of an is_null check after grabbing an index that just always exists. Also, as this currently isn't the same for all items, checking if an index exists seems like the right thing to do to me.

- The last option is to keep it all as it is right now. This would ensure backwards compatibility and add the flags.

Let me know what you think, before I start editing it all :)

PseudoKnight:
I never suggested changing the ENUM. I was lamenting that they start with HIDE_ in bukkit.

The fireworks meta is already being worked on and will hopefully include that change.

The enchants are kind of weird. As the oldest meta, it's the least sensible. It will probably remain untouched for now. I mean, that's certainly something I've brought up in the past, particularly with it appearing twice in the item array.

I've mentioned before that if it was up to me originally, I would have left out indexes with no values. But that's probably not something that we're going to change now without a good reason. It'd break too many scripts. So for consistency's sake, (unless others disagree?) I'd leave the index in and set it to null. If we want to break things, we can do it all in one go later.

PR #321, opened 2015-07-29
Add player weather functions

PseudoKnight:


src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
```diff
@@ -3432,6 +3433,113 @@ public Construct exec(Target t, Environment environment, Construct... args) thro
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0, 1};
+		}
+
+		@Override
+		public String docs() {
+			return "void {[player]} Returns true if the given player is experiencing precipitation.";
```
LadyCailin: Does this return true if the global weather is raining, or just if set_pstorm(true) has been previously called? This should be documented here to avoid confusion.

PseudoKnight: That wasn't clear in the Bukkit documentation either, so I matched their phrasing. I'll test it.

PseudoKnight: Turns out it only shows storm status for client side weather. (eg. that set by set_pstorm)

PR #325, opened 2015-08-14
Added set_pheld_slot and pheld_slot function.

Pieter12345:
void pset_selected_quickbar_slot([player], slot)
The function selects the specified quickbar slot.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
+++ b/src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
@@ -1814,11 +1814,11 @@ public class InventoryManagement {
 	}
 	
 	@api(environments = {CommandHelperEnvironment.class})
-	public static class pset_selected_quickbar_slot extends AbstractFunction {
+	public static class set_pheld_slot extends AbstractFunction {
 
 		@Override
 		public String getName() {
-			return "pset_selected_quickbar_slot";
+			return "set_pheld_slot";
 		}
 
 		@Override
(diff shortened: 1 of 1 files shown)


PseudoKnight:
set_pheld() or set_pslot() are two possibilities.

src/main/java/com/laytonsmith/core/functions/InventoryManagement.java
```diff
@@ -1812,6 +1812,88 @@ public CHVersion since() {
...
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{1};
+		}
+
+		@Override
+		public String docs() {
+			return "void {[player], slotNumber} Sets the selected quickbar slot of the given or executing player"
+					+ " to the given slot. The slots are in range of [1-9].";
```
LadyCailin: I have reservations doing 1 index. I think 0 index would be better here (slots 0-8) instead.

PseudoKnight: Agreed. All other functions/events that use slots start at 0.

Pieter12345: I did this because the slot numbers now correspond to the keyboard key you have to press to select it. I thought that was nice, but I can change it if desired.

LadyCailin: Yeah, let's change it, because we want to cater to the programmer, not the player. Since everywhere else in code is going to have references stored in 0-index terms, we don't want to change that for this one function. Not to mention, eww, 1-indexing.

PR #328, opened 2015-08-20
Fixes literal keywords in labels; Fixes CMDHELPER-2990 (null slot in arrays)

PseudoKnight:


PseudoKnight:
I think I need to add some functionality to support array('null': @item) for set_pinv temporarily, so it doesn't silently break for people all over again.

src/main/java/com/laytonsmith/core/constructs/CDouble.java
```diff
@@ -42,6 +42,6 @@ public CDouble clone() throws CloneNotSupportedException{
 
     @Override
     public boolean isDynamic() {
-        return true;
+        return false;
```
LadyCailin: What is the purpose of this change?

PR #332, opened 2015-08-25
Fix leading zeroes being trimmed after decimals

PseudoKnight:
I'd like to fix this in a more optimal way, but let's do this for now. It seems the least likely to have breaking unintentional side effects without reverting the "Fixed an issue where space mattered " commit.


LadyCailin:
Hmm, this is interesting. I think the correct solution may be to change it to a "provisional int" or something where it is stored as a string until the __autoconcat__ runs, at which point, depending on the context, will change it into a CInt or appropriately parse it as a CDouble. Good catch though!

PR #334, opened 2015-08-27
Better CLabel and CDouble handling in compiler

PseudoKnight:
Static.resolveConstruct() will mess up doubles. Currently this is causing -0.5 to turn into 0.5. So this can't be handled any later in the compiler unless we used some other construct resolver here. Putting checks for CDouble in the compiler instead of the lexer still addresses the whitespace issue you were fixing in the first place.

I also simplified CLabel creation by just using the check previously only used for CSlice.


LadyCailin:
Ah, crap. Ok, well if this works, then this will be fine. You know though, I'm not sure I added unit tests for the con at issue. Can you add a few unit tests then pull this? Namely, we need to test that 0 . 3 (note the spaces) is a double, and 0 . '3' and '0' . 3 are strings, and the same results if you remove the spaces.  If you add those and this passes, feel free to go ahead and pull it.

PseudoKnight:
Oh, you just reminded me of one or two more conditions when starting numbers with decimals. '0'.3 should be '03' not '0 0.3'. Should only be a double if the next token is a literal and the previous one isn't a string. Fixing that and adding tests.

PR #335, opened 2015-08-29
Pass bind() custom parameters by reference, not value

PseudoKnight:
This is an improved overwrite fix to the one in build 2927. This is how custom parameters functioned prior to my previous fix when using the assign() workaround. 

With this:
```
@a = array(1, 2, 3);
foreach(@b in @a) {
 // bind
}
```
I tested these custom parameters:
```
@b = @b // the old workaround
@a = 1 // this breaks the array reference outside the bind, just like prior to 2927
@b
@a
@c = @a[0];
```
I also incremented @a[0] inside the bind and outside the bind to verify the reference, in cases where it shouldn't have been passed and where it should have.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/EventBinding.java
+++ b/src/main/java/com/laytonsmith/core/functions/EventBinding.java
@@ -119,8 +119,7 @@ public class EventBinding {
 					throw new ConfigRuntimeException("The custom parameters must be ivariables", ExceptionType.CastException, t);
 				}
 				IVariable cur = (IVariable) var;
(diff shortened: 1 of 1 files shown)


LadyCailin:
FYI, moving forward, I would eventually like to implement bind so that it takes a closure, and doesn't accept extra parameters. Something like this:

bind('event_name', array(options: ...), array(prefilters: ...), closure(@event){});

This would solve all the problems that currently exist with binds, and would be much more elegant, both in terms of the mscript and java.

PR #340, opened 2015-09-21
Fixes CMDHELPER-3066 (error in core, ClassCastEx)

Pieter12345:
assign(datatype, notAnIvar, value) throwed a stacktrace because it
attempted to cast the notAnIvar to IVariable. This also happend when
using for example this code (as it gets optimized to assign()):
array NotAnIvar = value;

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/DataHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/DataHandling.java
@@ -234,6 +234,9 @@ public class DataHandling {
 			String name;
 			if(args.length == 3){
 				offset = 1;
+				if(!(args[offset + 0] instanceof IVariable)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/laytonsmith/core/functions/DataHandling.java
```diff
@@ -234,6 +234,9 @@ public Construct exec(Target t, Environment env, Construct... args) throws Cance
 			String name;
 			if(args.length == 3){
 				offset = 1;
+				if(!(args[offset + 0] instanceof IVariable)) {
+					throw new ConfigRuntimeException("assign with 3 arguments only accepts an ivariable as the second argument", ExceptionType.CastException, t);
```
LadyCailin: getName() + " with 3 arguments..."

Pieter12345: Changed that, also in the code that was already there.

PR #341, opened 2015-09-23
Made -@b and - @b the same and made @a=-@b work.

Pieter12345:
- Made -@b stay -@b instead of the string "-@b" (as - @b behaves). Also
works with $vars.
- Made notAVariable -@var compile as neg(@var). Also works with $vars.
- Fixes CMDHELPER-3104 and CMDHELPER-3102.
- Removed unused imports.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/StringHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/StringHandling.java
@@ -46,6 +46,7 @@ import java.util.HashSet;
 import java.util.IllegalFormatException;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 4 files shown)


LadyCailin:
2 things, does something like this work: msg(-add(1, 1)) (we expect -2) and can you add a unit test for this case? If you don't know how to do that, I'd be glad to discuss that on IRC.

src/main/java/com/laytonsmith/core/functions/StringHandling.java
```diff
@@ -1157,7 +1158,7 @@ public Construct exec(Target t, Environment environment, Construct... args) thro
 				params[i] = convertArgument(arg, c, i, t);
 			}
 			//Ok, done.
-			return new CString(String.format(formatString, params), t);
+			return new CString(String.format(Locale.US, formatString, params), t);
```
LadyCailin: Hmm. Let's discuss this.

Pieter12345: This fixes the failing unit test on all OS'es that use 1,23 as decimal seperator instead of 1.23.

LadyCailin: Yes, but I do not like changing it to US locale, without giving them the
option to change it. It's fine as a default, but not as a forced choice.

PR #353, opened 2015-11-04
CArray creation optimizations

PseudoKnight:
The array_get change consistently made a 1000x for() loop 7% faster.

When CArray normal arrays are created, and then later set with string index, it throws an exception, which can be costly to create. This was noticeable in profiling. I don't know the exact performance improvement from this change, but given how often this happens, it's probably significant. For example, it was throwing at least 1, and sometimes up to 3, exceptions for every event bind processed (not including exceptions thrown in the scripts).

I triple checked all the CArray(t, -1) to make sure they were appropriate and have been running these changes for a few days on my production server.

### DIFF
--- a/src/main/java/com/laytonsmith/core/ObjectGenerator.java
+++ b/src/main/java/com/laytonsmith/core/ObjectGenerator.java
@@ -747,7 +747,7 @@ public class ObjectGenerator {
 	 * @return
 	 */
 	public CArray color(MCColor color, Target t){
-		CArray ca = new CArray(t);
+		CArray ca = new CArray(t, -1);
 		ca.set("r", new CInt(color.getRed(), t), t);
 		ca.set("g", new CInt(color.getGreen(), t), t);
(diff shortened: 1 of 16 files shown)


LadyCailin:
Ideally you would just use CArray.GetAssociativeArray(). But that might need to be cleaned up a little to make it so that it fully utilizes your fix.

PseudoKnight:
I thought about that, but those methods seem a bit odd to me. One creates a normal array and then "forces" it to be associative. The other creates an associative array and then still forces it to be associative. But oddly the force associative mode method is basically just checking for null when the constructor always initializes it with a TreeMap.

LadyCailin:
Well, what I'm saying is, we should fix those methods, then use them.

PseudoKnight:
Should those methods just pass to the constructors, then? (without force)

LadyCailin:
Yes, I think that is acceptable.

PR #355, opened 2015-11-05
Added "/hug" example to register_command.

Pieter12345:
Who doesn't like hugs? The example shows how to use all indices of the
options array.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/Commands.java
+++ b/src/main/java/com/laytonsmith/core/functions/Commands.java
@@ -18,6 +18,7 @@ import com.laytonsmith.core.constructs.CVoid;
 import com.laytonsmith.core.constructs.Construct;
 import com.laytonsmith.core.constructs.Target;
 import com.laytonsmith.core.environments.Environment;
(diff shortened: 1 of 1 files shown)


src/main/java/com/laytonsmith/core/functions/Commands.java
```diff
@@ -276,6 +277,40 @@ public String docs() {
...
+		@Override
+		public ExampleScript[] examples() throws ConfigCompileException {
+			return new ExampleScript[]{
+				new ExampleScript("Register the /hug <player> command.",
+						"register_command('hug', array(\n"
+						+ "\t'description': 'Spread the love!',\n"
+						+ "\t'usage': '/hug <player>',\n"
+						+ "\t'permission': 'perms.hugs',\n"
+						+ "\t'noPermMsg': 'You do not have permission to give hugs to players (Sorry :o).',\n"
+						+ "\t'tabcompleter': closure(@alias, @sender, @args,\n"
```
LadyCailin: use the new syntax:

    closure(@alias, @sender, @args){
        ...
    }

PR #358, opened 2015-11-07
Re-added the old (now deprecated) CArray.push syntax.

Pieter12345:
Many extensions were not compatible with CH anymore. This commit adds a
final deprecated push that gives extension developers some time to
update their extensions.


LadyCailin:
Oops. I meant to do this anyways.

PR #361, opened 2015-11-16
Changed wrong return type of pspectator_target.

Pieter12345:
Changed VOID to UUID.


LadyCailin:
UUID isn't a data type. It's a string.

Pieter12345:
I went with UUID because puuid() states to return an UUID too.

PR #367, opened 2015-11-25
Add finalamount to entity damage events

PseudoKnight:


PseudoKnight:
Should this maybe be effectiveamount?

PseudoKnight:
Also, I just figured out that this was added in 1.7.9, so if we want backwards compatibility before that I'd need to add some checks. I'm not sure how practical backwards compatibility is for 1.7.2, though. It "works" but I've never seen anyone do any extensive testing.

PseudoKnight:
After sleeping on it (and helping someone today on 1.6.4), I think we should focus on backwards compatibility with 1.7.10 and leave 1.6.4 for backports. There's just too many issues and obstacles. Until we have a reliable and easy way to test builds on old versions of craftbukkit, it's just not very practical. It took me over 5 minutes just to find when they added this really small change, and the farther you go back the harder it gets now that bukkit repo is thrashed, extensions incompatible, builds and dependencies are gone.

The name, though... effectiveamount makes more sense, I think, but the only other person to chime in on the topic thought finalamount was good.

PseudoKnight:
Okay, no one else has an opinion. I'm good with this PR then.

PR #369, opened 2015-11-30
Feature: stringed arrays render strings correctly

lukecfairchild:
When stringing a array it will now wrap all strings inside the array with ticks and cancelling any ticks inside that string


LadyCailin:
This will cause probably a hundred test failures. If you want to do this, then that's fine, but all the tests need to be fixed to fix them as well. Additionally, we need to give people some lead time, in case this breaks backwards compatibility.

Pieter12345:
@LadyCailin I actually believe most tests are being done with integers. This commit only changes CString appearances in arrays, so tests will actually not be affected.
Example: msg(array('tes\'t')) --shows--> {'tes\'t'}

It will break code where arrays are compared to strings such as array('test') == '{test}'.

PR #371, opened 2015-12-06
Tab complete online player names by default for registered commands

PseudoKnight:
If you return null in onTabComplete, Bukkit will default to online player name tab completion. (players the sender can see) It also defaults to players if a tab completer isn't set, but we need to set a tab completer so that our event fires for all registered commands.


src/main/java/com/laytonsmith/core/events/drivers/ServerEvents.java
```diff
@@ -159,10 +163,7 @@ public String docs() {
 
 		@Override
 		public boolean matches(Map<String, Construct> prefilter, BindableEvent event) throws PrefilterNonMatchException {
-			if (event instanceof MCCommandTabCompleteEvent) {
-				return true;
-			}
-			return false;
+			return event instanceof MCCommandTabCompleteEvent;
```
jb-aero: Wow, that's embarrassing.

PseudoKnight: Hah. There's actually several of these. I fixed them when I was trying to improve prefilters, but I haven't had a reason to commit them. I usually just discard these types of changes, but this was related enough that I let it slide.

PR #372, opened 2015-12-09
Rewrite player_move and vehicle_move to use server events. Add world prefilters.

PseudoKnight:
This is a major internal change to how player_move and vehicle_move events are fired. The old method uses threading to reduce overhead, but it had low accuracy and was prone to ConcurrentModificationExceptions. (in some cases causing data corruption, though this could be significantly improved) This uses actual server events to avoid threading problems, but maintains most of the benefit of thresholds. But since thresholds are calculated on the main thread, it does increase overhead. And since it isn't throttled to 500ms, it can fire more frequently at a threshold of 1. (200-250ms when walking) That part is probably not that bad, because before it would fire all player_move events every 500ms, which could cause micro-stutter. Also, now if you cancel, it cancels the actual event, creating invisible barriers at the thresholds rather than sending the player to the threshold-from location. This also means that plugins that cancel the event later wouldn't be known to the script, much like other events.

In addition, I added an unbind() method to events so that I could clean up thresholds in lastPlayerLocations properly and keep that overhead down.

I'm currently running this on my live server, and while it takes the most overall time of all events (with two non-trivial binds at 1 and 10 thresholds) it doesn't take much time per tick.

There's a lot to discuss here. If you have any questions or concerns, please post them. Some of you may even think this is a non-starter and we should just fix the async event, which is fair. Personally I'd rather have something like this implementation, though it could probably be improved/optimized more.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitPlayerListener.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitPlayerListener.java
@@ -1,40 +1,24 @@
 package com.laytonsmith.abstraction.bukkit.events.drivers;
 
+import com.laytonsmith.abstraction.MCLocation;
+import com.laytonsmith.abstraction.bukkit.BukkitMCLocation;
 import com.laytonsmith.abstraction.bukkit.entities.BukkitMCPlayer;
-import com.laytonsmith.abstraction.bukkit.events.BukkitPlayerEvents;
+import com.laytonsmith.abstraction.bukkit.events.BukkitPlayerEvents.*;
 import com.laytonsmith.commandhelper.CommandHelperPlugin;
 import com.laytonsmith.core.Static;
 import com.laytonsmith.core.constructs.Target;
 import com.laytonsmith.core.events.Driver;
 import com.laytonsmith.core.events.EventUtils;
+import com.laytonsmith.core.events.drivers.PlayerEvents;
 import org.bukkit.Bukkit;
+import org.bukkit.Location;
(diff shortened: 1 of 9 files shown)


src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitPlayerListener.java
```diff
@@ -232,69 +220,104 @@ public void onPlayerTeleport(PlayerTeleportEvent event) {
 			return;
 		}
 		
-		BukkitPlayerEvents.BukkitMCPlayerTeleportEvent pte = new BukkitPlayerEvents.BukkitMCPlayerTeleportEvent(event);
+		BukkitMCPlayerTeleportEvent pte = new BukkitMCPlayerTeleportEvent(event);
 		EventUtils.TriggerListener(Driver.PLAYER_TELEPORT, "player_teleport", pte);
+		if(!pte.isCancelled()) {
```
LadyCailin: Should we maybe move this to a MONITOR level event handler?

LadyCailin:
If this seems performant, then I'm fine with this being pulled in general. The thresholds were a big part of what I had originally wanted to accomplish, (so we could keep as much processing out of mscript as possible) the rest was just kind of... whatever.

src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitPlayerListener.java
PseudoKnight: Ya, I agree. Doesn't seem to be a downside.

PR #376, opened 2015-12-16
Fixes

PseudoKnight:


src/main/java/com/laytonsmith/core/events/Prefilters.java
```diff
@@ -129,25 +146,17 @@ public static void match(Map<String, Construct> map, String key,
...
-            i2 = split[0].trim();
-        }
-        if(!i1.trim().equalsIgnoreCase(i2.trim())){
-            throw new PrefilterNonMatchException();
-        }
-    }
+	private static void ItemMatch(Construct item1, Construct item2) throws PrefilterNonMatchException{
+		String i1 = item1.val().split(":")[0];
+		String i2 = item2.val().split(":")[0];
+		if (!i1.trim().equals(i2)) {
```
LadyCailin: i2.trim()

LadyCailin:
After fixing that .trim(), feel free to pull this at your leisure.

src/main/java/com/laytonsmith/core/events/Prefilters.java
PseudoKnight: i2 is always generated by the event, so I figured it didn't need trimming. Do I still need to trim it?

PR #381, opened 2015-12-25
Replace deprecated import/export format with array keys and default values

PseudoKnight:
These are the changes mentioned here:
http://dev.enginehub.org/youtrack/issue/CMDHELPER-2847

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/DataHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/DataHandling.java
@@ -2703,7 +2703,7 @@ public class DataHandling {
 
 	@api
 	@seealso({_export.class})
-	public static class _import extends AbstractFunction implements Optimizable {
+	public static class _import extends AbstractFunction {
 
 		@Override
 		public String getName() {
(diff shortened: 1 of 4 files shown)


src/main/java/com/laytonsmith/core/functions/DataHandling.java
```diff
@@ -2815,25 +2780,22 @@ public String getName() {
...
-					+ " the value is already stored, it is overwritten. See {{function|import}} and"
-					+ " [[CommandHelper/import-export|importing/exporting]]. The reference to the value"
-					+ " is stored, not a copy of the value, so in the case of arrays, manipulating the"
-					+ " contents of the array will manipulate the stored value. export() is threadsafe.";
+			return "void {key, value} Stores a value in the global storage register."
+					+ " An arbitrary value is stored with the given key, and can be retreived using import."
+					+ " If the value is already stored, it is overwritten. See {{function|import}}."
+					+ " The reference to the value is stored, not a copy of the value, so in the case of"
+					+ " arrays, manipulating the contents of the array will manipulate the stored value."
+					+ " export() is threadsafe.";
```
LadyCailin: We should add a blurb about the array('key1', 'key2') format in the docs, as well as probably add an example. This is true for both functions.

PseudoKnight: I already added an example, but I'll add a note in the description.

PseudoKnight: How's this?

```
return "mixed {key, [default]} This function imports a value from the global value register. It looks for a"
		+ " value stored with the specified key (using the export function), and returns that value."
		+ " If specified key doesn't exist, it will return either null or the default value if specified."
		+ " An array may be used as a key. It is converted into a string with the array values separated by"
		+ " dots. import() is threadsafe.";
```

LadyCailin: Sounds good!

PR #383, opened 2015-12-29
Fixed a few errors in core by max_item_stack.

Pieter12345:
- Fixed multiple errors in core caused by invalid itemType formats
  ("123:").
- Added more specific exceptions.
- Improved performance: Removed splits and the need to clone the passed
  array.

### DIFF
--- a/src/main/java/com/laytonsmith/core/ObjectGenerator.java
+++ b/src/main/java/com/laytonsmith/core/ObjectGenerator.java
@@ -259,6 +259,7 @@ public class ObjectGenerator {
      * new formats currently
      *
      * @param i
+	 * @param t
      * @return
      */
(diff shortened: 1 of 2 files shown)


PseudoKnight:
Doing a String.matches(), String.indexOf() and String.substring() is faster than String.split()?

Pieter12345:
I believe String.indexOf() and String.subString() is, the String.matches() is added to give more specific exceptions and as a whole, it fixes an error in core when the item type would be "0:".

PseudoKnight:
This PR was out-dated and I wanted to do some of it differently, so I rewrote these fixes. The most notable thing I kept was the "seperatorIndex", which was a good strategy, though I slightly improved that as well.

PR #384, opened 2016-01-01
Breaking extension updates

PseudoKnight:


PseudoKnight:
This would be one of those times in theory we could bump the version. At the very least we should have one commit that says "BREAKING CHANGES FOR EXTENSIONS". I'll be updating my Update Guide, but there's little guarantee people see that.

LadyCailin:
Oh man, a version bump. I'm not sure I'm ready for that kind of commitment.

Just kidding \^_\^ That actually sounds like a pretty good idea. I'll do that after I get all this merged in. Anybody want to get a changelog together? lol.

PseudoKnight:
Heh. If you're not committed to versions, they won't help any. But if you're consistent they can be a decent way of communicating things. Don't do it if you don't have a plan for when you'll change which version numbers.

I've gone through over half of the working extensions so far and updated 11 of those that needed updating. I can push those changes when we're ready to go. I'll probably bump their versions too.

LadyCailin:
It will be useful to go ahead and bump the CH version. I still am mostly going to use build numbers for normal development, but it will be a useful reference to be able to say "this extension version is only compatible with 3.3.2 of CH." plus, a useful tag in the commits.

jb-aero:
Hi, I try to check Github at least once a day. So far my boss hasn't asked me to work on weekdays, so I can even do stuff, like update my own extensions :)

PR #339, opened 2015-09-19
Update to ASM 5.0

Techcable:
Supports Minecraft Servers that use Java 8
Needed for [TacoSpigot](https://github.com/TacoSpigot/TacoSpigot)

PR #386, opened 2016-01-12
Cleanup classfile reading and upgrade to latest ASM

Techcable:
Fixes compatibility with PaperSpigot and TacoSpigot.
You were using a very old version of ASM, which can't read java 8 jars.
Since PaperSpigot updated to 1.8, you need to update ASM in order to support them.

Unlike #339 I can confirm that metadata-caching works.
All unit-tests pass, but I don't have scripts to test with.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -336,6 +336,12 @@
 			<groupId>org.java.net.substance</groupId>
 			<artifactId>substance</artifactId>
 			<version>6.0</version>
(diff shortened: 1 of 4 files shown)


Techcable:
Uh oh:
`[12:09:54 INFO]: [CommandHelper] Running initial class discovery, this will probably take a few seconds...`

PR #387, opened 2016-01-15
Extension breaking changes and removal of deprecations

PseudoKnight:
Also included is a core error fix when handling enchantment arrays with invalid enchantment names.


PseudoKnight:
Welp, now I have to resubmit this.

LadyCailin:
You could have just added a new commit. :D

PseudoKnight:
I like clean commit histories. =P

Though I think I could have amended the commit? I haven't tried that with a PR.

PR #388, opened 2016-01-21
Breaking changes and some fixes

PseudoKnight:
Re-submitted to resolve new merge conflicts.


PseudoKnight:
Compile failed because of two uses of CArray.push(construct) that you added.

PR #393, opened 2016-02-02
This bug broke all extensions. Click here if you want to see the astonishing fix.

Pieter12345:
Having extensions caused NullPointerExceptions when events fired. This
has been fixed in a very clean way.


LadyCailin:
ಠ_ಠ

PR #395, opened 2016-02-10
Don't schedule tasks when the plugin is disabled.

Pieter12345:
Queue, thread unsafe and x_run_on_main_thread_... function calls caused
an Exception to be thrown by the Bukkit scheduler when the server was
shutting down. This PR hopefully fixes that.


src/main/java/com/laytonsmith/abstraction/bukkit/BukkitConvertor.java
```diff
@@ -476,6 +477,9 @@ public MCInventory GetLocationInventory(MCLocation location) {
 
 	@Override
 	public void runOnMainThreadLater(DaemonManager dm, final Runnable r) {
+		if(!CommandHelperPlugin.self.isEnabled()) {
+			throw new CancelCommandException("CommandHelper tried to schedule a task while the plugin was disabled (is the server shutting down?).", Target.UNKNOWN);
```
LadyCailin: Use Implementation.GetServerType().getBranding() instead of hard coding the string "CommandHelper". This applies in both places.

PR #396, opened 2016-02-14
Check format before modifying the player_chat event.

Pieter12345:
modify_event in player_chat would throw Exceptions that were displayed
in console as:
- FormatException: Conversion = '"'
- FormatException: d !=
org.bukkit.craftbukkit.v1_8_R3.entity.CraftPlayer
- Errors in core

This commit checks if the new format is valid for how MineCraft
interpretes it and gives a better Exception message.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
@@ -43,13 +43,16 @@ import com.laytonsmith.core.events.drivers.EntityEvents.entity_death;
 import com.laytonsmith.core.exceptions.CRE.CREBindException;
 import com.laytonsmith.core.exceptions.CRE.CRECastException;
 import com.laytonsmith.core.exceptions.CRE.CREFormatException;
+import com.laytonsmith.core.exceptions.CRE.CRENullPointerException;
 import com.laytonsmith.core.exceptions.ConfigRuntimeException;
 import com.laytonsmith.core.exceptions.EventException;
 import com.laytonsmith.core.exceptions.PrefilterNonMatchException;
+import com.laytonsmith.core.functions.EventBinding.modify_event;
 import com.laytonsmith.core.functions.StringHandling;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.FormatFlagsConversionMismatchException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IllegalFormatConversionException;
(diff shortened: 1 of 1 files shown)


PseudoKnight:
Oops, try again.

Pieter12345:
@PseudoKnight I should really move away from this GitHub Windows client. It wanted me to merge things after a rebase and when I did that, it didn't notice it at all. I managed to force everything into one commit now though.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -1537,9 +1540,17 @@ public boolean modifyEvent(String key, Construct value, BindableEvent event) {
                     }
                 }
 				if("format".equals(key)){
+					String format = value.nval();
+					if(format == null) {
+						throw new CRENullPointerException("The \"format\" key in " + modify_event.class.getSimpleName() + " for the " + this.getName()
```
LadyCailin: new modify_event().getName(). It's not based on the class name.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -1537,9 +1540,17 @@ public boolean modifyEvent(String key, Construct value, BindableEvent event) {
...
+								+ " event may not be null.", Target.UNKNOWN);
+					} else if(format.replaceAll("%%", "").replaceAll("\\%\\%|\\%[12]\\$s", "").contains("%")) {
+						throw new CREFormatException("The \"format\" key in " + modify_event.class.getSimpleName() + " for the " + this.getName()
+								+ " event only accepts %1$s and %1$s as format specifiers. Use a \"%%\" to display a single \"%\".", Target.UNKNOWN);
```
PseudoKnight: "accepts %1$s and %1$s" are the same.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -1537,9 +1540,17 @@ public boolean modifyEvent(String key, Construct value, BindableEvent event) {
                     }
                 }
 				if("format".equals(key)){
+					String format = value.nval();
+					if(format == null) {
+						throw new CRENullPointerException("The \"format\" key in " + modify_event.class.getSimpleName() + " for the " + this.getName()
+								+ " event may not be null.", Target.UNKNOWN);
+					} else if(format.replaceAll("%%", "").replaceAll("\\%\\%|\\%[12]\\$s", "").contains("%")) {
```
PseudoKnight: Seems inefficient to run this on every chat event. This can be combined into one precompiled pattern. Maybe it should be a compile time check?

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
Pieter12345: Will be changed in the next commit.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
Pieter12345: Since the ..Event.setFormat(String format) throws an exception when the format is invalid, I moved the regex to the catch so that it will only be used to generate a better exception and not run on valid use.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
Pieter12345: Whoops, thanks.

PR #398, opened 2016-03-01
Update Biomes, Entities, DamageCauses, Materials, and Inventory slots for 1.9

PseudoKnight:
Using -106 for the numerical player slot to match [Minecraft](http://minecraft.gamepedia.com/Player.dat_format#Inventory_slot_numbers). This is the index stored in NBT. I know it's weird. Our alternatives are probably -1, 99, and 104, if someone wants to make a case for it.


PseudoKnight:
The other thing that we could change is off_hand for the mob equipment slot. We could lose the underscore.

PseudoKnight:
One odd concession I had to make was 'splashpotion', which has its own ID now. The data values aren't used in 1.9, but I left those all in for backwards compatibility and just changed 'splashpotion' to the new ID.

LadyCailin:
I think -106 is fine.

PR #401, opened 2016-03-15
More 1.9 updates

PseudoKnight:
New potion array:
{name: SPLASH_POTION, data: 0, meta: { base: {type: REGEN, extended: true, upgraded: false}, potions: {}}}
This replaces the old data values that don't work anymore.

New item_swap event for when players swap their main hand and off hand items using the "f" key (by default).

Also a couple of missing ENUM types from the latest updates.


PseudoKnight:
Thought: is item_swap sufficiently explanatory? Technically you can "swap" items in the inventory_click event, but this is different. However, item_held_swap sounds a bit awkward; item_hand_swap sounds like you're swapping hands (a different action); item_in_hands_swap... swap_items_in_hand... swap_hand_items?

LadyCailin:
item_swap is fine

PR #406, opened 2016-03-19
http_request charset fix

itstake:


LadyCailin:
What use case does this fix? Reading sites with higher level characters?

PR #408, opened 2016-03-22
Added block_from_to event

itstake:
Added block_from_to event


PseudoKnight:
No. One of those prefilters should probably be removed. Block "type" and "data" should probably be CInt instead of CString. The documentation mentions xyz under the "block" index, which doesn't exist.

This event can be no joke when it comes to frequency sometimes, so be careful.

jb-aero:
Should probably also be using material names instead of numeric ids.

PR #410, opened 2016-03-30
Added info about range in RangeException in spawn_mob.

Pieter12345:
- Added the max qty to the docs.
- Added the max qty in the RangeException that is thrown when the max
qty is exceeded.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/Minecraft.java
+++ b/src/main/java/com/laytonsmith/core/functions/Minecraft.java
@@ -409,8 +409,8 @@ public class Minecraft {
 			if (args.length > 1) {
 				qty = Static.getInt32(args[1], t);
 				if (qty > spawn_mob.SPAWN_LIMIT) {
-					throw new CRERangeException("A bit excessive to spawn more than " + spawn_mob.SPAWN_LIMIT
-							+ " mobs, don't you think? Let's scale that back some, huh?", t);
+					throw new CRERangeException("You can not spawn more than " + spawn_mob.SPAWN_LIMIT
+							+ " mobs at once using the " + this.getName() + " function.", t);
 				}
 			}
 			MCLocation l;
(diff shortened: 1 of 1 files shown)


PseudoKnight:
This is good, but maybe we should just remove the limit? If we're trying to save people from killing the server, there's a ton of ways they can do that. (unless there's a limit in bukkit)

Pieter12345:
@PseudoKnight I agree. I could look if Bukkit has a limit later today.

jb-aero:
I'm pretty sure it doesn't unless it is specifically on the method this uses. spawn_entity can do the same thing and doesn't have a limit.

If you are keeping this limit though, I'd put the cap in a constant and use that instead of a literal 50 in these various locations.

PseudoKnight:
If we do use a cap, something like 1000 would make more sense. You'd freeze a smaller server for a while with that amount and it would even cause problems on the beefiest of machines. The idea is we're not in charge of performance on servers, but we can prevent someone from spawning 100000 entities at once and crashing the server and requiring external tools to fix.

Pieter12345:
@jb-aero I put the spawn limit in a variable.
@PseudoKnight I've upped the spawn limit to 10.000.

Tests gave me the following result:
- 100 mobs were spawned in 18ms, 14ms, 20ms, 18ms.
- 1000 mobs were spawned in 261ms, 216ms, 219ms, 236ms.
- 10000 mobs were spawned in 1753ms, 1715ms, 1647ms, 1561ms.

Used code to test: @time = time(); spawn_mob('OCELOT:BLACK_CAT', 100, ploc('woesh0007')); broadcast('Time elapsed: '.(time() - @time));

Spawning 10.000 mobs did not disconnect me yet, but I'm sure it would on some servers. I don't think anyone would ever want to go past this limit within 1 server tick.

LadyCailin:
Let's change the error message to sound more... professional :D

Pieter12345:
Oh, who who would not like to have a 10.000 cat-disk on their server? It does make my client and server use 15% CPU though.
[http://imgur.com/lJKLAFA](url)

Pieter12345:
@LadyCailin I thought it might have been one of those things that should not be touched. Thank you for saying that, I changed it.

PR #412, opened 2016-04-02
Fixed an NPE in mod(@var, 0).

Pieter12345:
Using mod(@var, 0) would throw a ArithmeticException ("Division by 0").
This commit fixes that.


src/main/java/com/laytonsmith/core/functions/Math.java
```diff
@@ -382,17 +382,20 @@ public String getName() {
 		public Construct exec(Target t, Environment env, Construct... args) throws CancelCommandException, ConfigRuntimeException {
 			long arg1 = Static.getInt(args[0], t);
 			long arg2 = Static.getInt(args[1], t);
+			if(arg2 == 0) {
+				throw new CRERangeException("Moduloing by 0!", t);
```
EntityReborn: Shouldn't this say "Modulo by 0" instead? Current phrasing seems awkward.

Pieter12345: Yeah, that'd be better. Thanks.

PR #413, opened 2016-04-06
Register item_swap and entity_toggle_glide dynamically.

jb-aero:
Also change registerEventsDynamic to be able to handle listeners with only the standard Bukkit annotation.


jb-aero:
Using the new dynamic registration, both old events only annotated with @ EventHandler and new events annotated with @ EventIdentifier are available and responsive ingame. The benefit to this is not having to modify every handler match the dynamic form.

PseudoKnight:
I like it, assuming it's working.

Fixes http://dev.enginehub.org/youtrack/issue/CMDHELPER-3131

jb-aero:
I tested with inventory_click and item_swap on 1.8 and 1.9. There were no errors in either version, both fired when expected on 1.9, and inventory_click fired as usual on 1.8.

PR #416, opened 2016-04-15
Added backwards compatibility to player_interact.

Pieter12345:
Since the getHand() method in PlayerInteractEvent does not exist in
Bukkit versions prior to 1.9, it became incompatible with those version
after this was added to CH. This fixes that.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
@@ -860,18 +860,28 @@ public class PlayerEvents {
         public boolean matches(Map<String, Construct> prefilter, BindableEvent e) throws PrefilterNonMatchException {
             if(e instanceof MCPlayerInteractEvent){
                 MCPlayerInteractEvent pie = (MCPlayerInteractEvent)e;
-                MCEquipmentSlot h = pie.getHand();
                 CString hand;
+				MCEquipmentSlot h;
+				
+				try {
+					h = pie.getHand();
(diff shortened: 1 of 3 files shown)


src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -860,18 +860,28 @@ public Driver driver() {
...
                 if(((MCPlayerInteractEvent)e).getAction().equals(MCAction.PHYSICAL)){
                     return false;
                 }
                 
-                if(h == MCEquipmentSlot.WEAPON) {
-                   hand = new CString("main_hand", Target.UNKNOWN);
-                } else {
-                    hand = new CString("off_hand", Target.UNKNOWN);
-                }
+				if(h == null) {
```
jb-aero: this looks like a typo

Pieter12345: @jb-aero I used TAB's, some parts of the file use whitespaces. When you look at the full file, it should make sense. Please let me know if you mean something else.

jb-aero: A "typo" usually refers to typing one word when you meant another, in this case checking if something is not null ONLY after a previous check says it is null.

I am aware that most older files are mostly spaces instead of tabs :P

Pieter12345: I see what happend there. Thank you for pointing it out. It has been fixed in the PR after this one :)

PR #418, opened 2016-04-19
Added in_cmdline_mode.

Pieter12345:
It'd be nice to have a function that tells you whether you are in
CMDline mode or not. This does that.


LadyCailin:
Hmm. It might be better to make a function that returns the current platform instead, I.e. SHELL or BUKKIT. Thoughts?

PR #425, opened 2016-05-22
Added support for snowman "derp mode" in the entity spec functions

macjuul:
"derp mode" is true if the snowman does't have a pumpkin on it's head


PseudoKnight:
This should have checked for 1.9.4, not 1.9, since it was only added 5 days ago.

PR #426, opened 2016-05-22
Added 'collisionrule' and 'deathmessagevisibility' team options

macjuul:
Only for MC 1.9 and up


macjuul:
Made another new addition, making a new PR for both changes

PR #427, opened 2016-05-22
'collisionrule' and 'deathmessagevisibility' options for teams & entity ai functions

macjuul:


PseudoKnight:
Setting the AI was added March 26th, a few weeks after 1.9 (right before 1.9.2). It might be a good candidate have checks in BukkitMCLivingEntity if the Bukkit methods exist. Alternatively, we could check for 1.9.2 due to its close proximity and guarantee of compatibility. I'm not sure which is faster, but I've seen LadyCailin say checking if the method exists is "correct". In those cases, I would add a comment in which version it was added, though.

macjuul:
Yes, I'll work on that soon. I also noticed the nametagvisibility still uses the old deprecated methods, so we might want to change that to use the same enums as these new options

PseudoKnight:
Yes, as long as it continues to work on older versions.

macjuul:
Do we need backwards compatibility for the old nametagvisibility values in the new builds as well?

macjuul:
I'll also added version checking to the glide and glow methods

PseudoKnight:
I can't support the last commit as it would break compatibility with scripts with almost no benefit. The rest looks good, though.

macjuul:
Sorry I've been rather busy lately. I'll be fixing it up and probably submitting it in a new PR when I have the time

macjuul:
Improved PR coming asap

PR #430, opened 2016-06-20
Allow dynamic array return in default switch case.

Pieter12345:
switch() with a dynamic array as default return value would give an
compile error about cases having to be static. This was caused by
optimization code handling the default code as case code.
Example of what used to give an error:
switch('bla', array(array()))
switch('bla', 'case1', 'return1', array(array('Default case return in 2
arrays')))

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
+++ b/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
@@ -634,54 +634,55 @@ public class BasicLogic {
 					}
 				}
 			});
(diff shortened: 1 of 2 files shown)


LadyCailin:
This is done intentionally. Switch statements can be optimized more heavily than if else chains, *because* the cases are constant. The cases are sorted into a jump table at compile time, which makes them much faster, making switches go from O(n) to O(log n). So, I'm going to decline this PR.

LadyCailin:
Oh. I misread this. Nevermind.

Pieter12345:
@LadyCailin Haha, I understand the confusion. All I did was make it not throw an Exception about cases having to be static when the default code (not case) is dynamic. This makes the code in the default case behave like code in any other case (as expected).

PR #434, opened 2016-07-29
New 'contains' and 'containsic' keywords

macjuul:
These are essentially aliases to the array_contains() and
array_contains_ic() functions.

**Example:** `array('one', 'two', 'three') contains 'two'; // True`

I know it's not a very needed addition, although in my opinion, these keywords do make code look quite a bit neater.


LadyCailin:
I'm against adding this for now. This problem will be solved once objects are a thing, and then this will be a obscure feature that has to be supported. Eventually, this will work: `@array->contains(1)` but any type of object will support that notation.

macjuul:
Having OOP would indeed be way better ;) Thanks anyway!

PR #435, opened 2016-07-29
Added 'collisionrule' and 'deathmessagevisibility' options for teams & entity ai functions

macjuul:
The NameTagVisibility *should* also use this new Option system that the new team options use, although I have no idea how to make it backwards compatible right now, so I left it out of the change for now

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCEntity.java
@@ -323,17 +323,12 @@ public class BukkitMCEntity extends BukkitMCMetadatable implements MCEntity {
 	
 	@Override
 	public boolean isGlowing() {
-		if(ReflectionUtils.hasMethod(e.getClass(), "isGlowing", null)) { // Added around 1.9
-			return e.isGlowing();
-		}
-		return false;
+		return e.isGlowing();
 	}
 	
 	@Override
 	public void setGlowing(Boolean glow) {
-		if(ReflectionUtils.hasMethod(e.getClass(), "setGlowing", null)) { // Added around 1.9
-			e.setGlowing(glow);
-		}
+		e.setGlowing(glow);
 	}
 
 	@Override
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCEntity.java
@@ -1,6 +1,7 @@
 package com.laytonsmith.abstraction.bukkit.entities;
 
 import com.laytonsmith.PureUtilities.Vector3D;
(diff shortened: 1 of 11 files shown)


src/main/java/com/laytonsmith/core/functions/Scoreboards.java
```diff
@@ -412,11 +414,16 @@ public Construct exec(Target t, Environment environment,
 				if(Static.getServer().getMinecraftVersion().gte(MCVersion.MC1_8)) {
 					ops.set("nametagvisibility", new CString(team.getNameTagVisibility().name(), t), t);
 				}
+				if(Static.getServer().getMinecraftVersion().gte(MCVersion.MC1_9_4)) {
```
PseudoKnight: This should be MCVersion.MC1_9, like in the other function.

src/main/java/com/laytonsmith/abstraction/enums/bukkit/BukkitMCOptionStatus.java
```diff
@@ -0,0 +1,36 @@
...
+	public static BukkitMCOptionStatus getConvertor() {
+		if (instance == null) {
+			instance = new BukkitMCOptionStatus();
+		}
+		return instance;
+	}
+
+	@Override
+	protected OptionStatus getConcreteEnumCustom(MCOptionStatus abstracted) {
+		if(Static.getServer().getMinecraftVersion().lt(MCVersion.MC1_9_4)) {
```
PseudoKnight: Added in MC1_9. Check the others too. It was within a few days of the initial release. No one should be using first week builds, so 1_9 is fine.

src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCEntity.java
```diff
@@ -322,12 +323,17 @@ public MCLocation asyncGetLocation() {
...
+		if(ReflectionUtils.hasMethod(e.getClass(), "isGlowing", null)) { // Added around 1.9
+			return e.isGlowing();
+		}
+		return false;
 	}
 	
 	@Override
 	public void setGlowing(Boolean glow) {
-		e.setGlowing(glow);
+		if(ReflectionUtils.hasMethod(e.getClass(), "setGlowing", null)) { // Added around 1.9
```
PseudoKnight: These were added on day 1 of 1.9 builds, so no check is required here.

PseudoKnight:
So, three version checks should be changed to 1_9, and the two method checks for glowing can be removed.

macjuul:
I'll fix it asap.

Also, what is the best way to find out when something was added or changed?

PseudoKnight:
I refer to Bukkit commit history here.
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits

PseudoKnight:
This looks good to me.

PR #436, opened 2016-09-15
Fix attempt for rare NPE in CClosure.execute().

Pieter12345:
This commit is an attempt to fix a very rarely occuring NullPointerException in CClosure.execute() running on a seperate thread (x_new_thread()) by synchronizing access to GlovalEnv.GetStackTraceManager() so it should never return null again.


Pieter12345:
I can't seem to update my branch to the CH:master branch for some reason (using git pull upstream master -> git push origin, it tells me I'm up to date). If you have any clue how to synchronize commits and remove those 3 older commits, please comment :)

PR #437, opened 2016-09-16
Added synchronize function.

Pieter12345:
This function synchronizes the code ran in the closure with any other
thread that has called the synchronize(name, closure) function using the
same name String (by value). For more info, read _synchronized.docs().

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/Threading.java
+++ b/src/main/java/com/laytonsmith/core/functions/Threading.java
@@ -21,6 +21,7 @@ import com.laytonsmith.core.constructs.Target;
 import com.laytonsmith.core.environments.Environment;
 import com.laytonsmith.core.environments.GlobalEnv;
 import com.laytonsmith.core.exceptions.CRE.CRECastException;
+import com.laytonsmith.core.exceptions.CRE.CREInsufficientArgumentsException;
 import com.laytonsmith.core.exceptions.CRE.CREThrowable;
 import com.laytonsmith.core.exceptions.CancelCommandException;
 import com.laytonsmith.core.exceptions.ConfigCompileException;
@@ -28,6 +29,8 @@ import com.laytonsmith.core.exceptions.ConfigRuntimeException;
 import com.laytonsmith.core.exceptions.FunctionReturnException;
 import com.laytonsmith.core.exceptions.LoopManipulationException;
 import com.laytonsmith.core.exceptions.ProgramFlowManipulationException;
+import java.util.HashMap;
+import java.util.Map.Entry;
 import java.util.concurrent.Callable;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +332,123 @@ public Version since() {
...
+					ret = CVoid.VOID;
+				} catch(FunctionReturnException e) {
+					ret = e.getReturn();
+				} catch(ConfigRuntimeException | ProgramFlowManipulationException e) {
+					ex = e;
+				}
+			}
+			
+			// Remove 1 from the call count or remove the synchronize object if there's no thread waiting for it.
+			synchronized(syncObjectMap) {
```
LadyCailin: This should happen in a finally block after the closure is executed. Otherwise, errors in one thread will cause other threads that call synchronize to never unlock.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +332,123 @@ public Version since() {
...
+							break searchLabel;
+						}
+					}
+					syncObjectMap.put(syncName, 1);
+					syncStr = syncName;
+				}
+			}
+			
+			// Get the closure arguments.
+			Construct[] closureArgs;
```
LadyCailin: Once this is an implicit closure, none of this code will be necessary.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +332,123 @@ public Version since() {
...
+		}
+		
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{Integer.MAX_VALUE};
+		}
+		
+		@Override
+		public String docs() {
+			return "mixed {string, closure} Synchronizes the code in the closure for all calls with the same string argument."
```
LadyCailin: This needs far more details. Synchronization is not a simple topic, so we need to say a bit more than this, I think. Also, examples should be added.

Pieter12345:
@LadyCailin Thanks for the feedback, I do agree with most of it. What is the use of synchronizing Thread.currentThread() though? Unless other threads can get access to this Thread object, it will not do anything.

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345: @LadyCailin In the code block I can just see one line of the docs String. Did you also read the 4 lines below this?

LadyCailin:
Oh, correct you are. It actually should synchronize with the parent class object, or the instance, if run from an instance method, but neither of those concepts exist yet, so don't worry about that. :D

src/main/java/com/laytonsmith/core/functions/Threading.java
LadyCailin: Yes, I read that, but there's not enough information about how or why one would need this, etc. if you like, you can just add links to external sources (I.e. Wikipedia) that goes into more detail about the theory.

Pieter12345:
Okay, I will make it synchronize based on Construct.val() String values, except for CArrays, which reference will be used (which is convenient since well, synchronizing an array and then modifying it in that sync block should still cause other threads to block when passing the same array reference to synchronized).

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +336,137 @@ public Version since() {
...
+									break;
+								}
+							}
+						}
+					}
+				}
+			}
+			
+			// Return the Construct received from evaluating the passed code block.
+			return ret;
```
LadyCailin: I see no reason to not return void

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +336,137 @@ public Version since() {
 
 	}
 
+	@api
+	@noboilerplate
+	@seealso({x_new_thread.class})
+	public static class _synchronized extends AbstractFunction {
+		private static final HashMap<Object, Integer> syncObjectMap = new HashMap<Object, Integer>();
```
LadyCailin: Use Map on the left side. Also, a TreeMap might be a better choice here, though actually, it probably won't matter, since the usual case will be 0 or 1 objects in here. Anyways, I would need to research it more, but you might need to use the _volatile_ keyword here. Or maybe that just applied to primitives. Anyways, go look that up.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -329,4 +336,137 @@ public Version since() {
...
+					+ " https://en.wikipedia.org/wiki/Synchronization_(computer_science)";
+		}
+		
+		@Override
+		public Version since() {
+			return CHVersion.V3_3_2;
+		}
+		
+	}
+
```
LadyCailin: Need examples

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345: HashMap VS TreeMap is O(1) VS O(log(n)) for get and put. I just went with a HashMap because there will rarely be more than 5 objects in there, I believe lookup speed won't be noticable anyways (the only way to compare which one is actually faster is by testing according to "google"). I will change the syncObjectMap to "Map".
The volatile keyword in java is for Object references that might be changed in between 2 calls by some other thread. In this case, the variable will never change (it's even declared final) so the volatile keyword does not apply (it's actually a compile error to use volatile with final).

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345: I left it in because I thought this syntax could be nice to use since it matches up with if():
`@var = synchronized(@array) { if(array_size(@array) > 1) { @array[0] } else { null } }`
I guess this could be seen as a feature or as bad programming, what do you prefer?

LadyCailin:
Let's just return void for now. We can always change it to return the value later, without breaking anybody's code, but not vice versa.

LadyCailin:
Merged. Thanks :)

PR #440, opened 2016-11-06
Don't trigger a player_teleport event if it was caused by an NPC

Murreey:


PseudoKnight:
"instanceof NPC" is unnecessary as those would not trigger player_teleport.

Murreey:
@PseudoKnight Oh, duh. Have removed it now.

PseudoKnight:
I'm not sure this is something we should add in CH directly since it can be done in scripts when desired. Also, apparently Citizens adds metadata on a delay, so it wouldn't work in some cases. Check the edit [here](http://dev.enginehub.org/youtrack/issue/WORLDGUARD-3749).

Murreey:
You can work around it with scripts but surely it'd be better for only players to trigger the player teleport event, as that's bound to trip a lot of people up. I guess it could also be solved in Spigot assuming that's related.

And the metadata seems to be fine (or at least gets added before they trigger the teleport event). It's not triggering even though Citizens loads after CH, or with a /citizens reload.

LadyCailin:
I think this is fine. "player" does imply non-npc, so.. Further thoughts?

jb-aero:
"Further thoughts": doesn't need the NPC import.

PseudoKnight:
Another thought: What other events do NPCs trigger?

Murreey:
Far as I can tell they also trigger world_changed if you TP them to a different world. Nothing of the other obvious ones like player_move or player_spawn trigger, but I'm not sure if there's a way to make them fish or use beds or anything more obscure.

They also trigger the region_change in CHWorldGuard when teleported, as it uses onPlayerTeleport too. That'll need fixed separately I guess.

PseudoKnight:
The number of events player NPCs affect is significant. I was working my way through them, but feel like maybe there should be some sort of singular way to fix this in the bound event triggering. Anyway, I'm trying to resolve old PRs, so to the tracker this goes.

http://dev.enginehub.org/youtrack/issue/CMDHELPER-3140

PR #442, opened 2016-11-24
Reduce noise in mostly harmless ENUM mismatch warnings

PseudoKnight:
The current ENUM converter checker takes an enormous amount of space per warning. This is mostly redundant information and is quite spammy on non-latest Spigot versions. I significantly reduced it by almost 200 characters, and it starts the message with the relevant ENUM type. If you still feel we need some messages about what causes this and if it can cause problems, I think that needs to be added before or after the checks.


LadyCailin:
Yeah, this is probably ok.

PR #443, opened 2016-12-06
Add equals_item function

EntryPointKR:
I want to use the CommandHelper more conveniently

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
+++ b/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
@@ -1333,7 +1333,7 @@ public class BasicLogic {
     }
 
     @api
-	public static class equals_item extends AbstractFunction {
+	public static class item_equals extends AbstractFunction {
 		@Override
 		public Version since() {
 			return CHVersion.V3_3_2;
@@ -1341,7 +1341,7 @@ public class BasicLogic {
 
 		@Override
 		public String getName() {
-			return "equals_item";
+			return "item_equals";
 		}
 
 		@Override
--- a/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
+++ b/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
@@ -1,14 +1,16 @@
 package com.laytonsmith.core.functions;
 
 import com.laytonsmith.PureUtilities.Version;
-import com.laytonsmith.abstraction.MCItemStack;
+import com.laytonsmith.abstraction.Convertor;
+import com.laytonsmith.abstraction.StaticLayer;
+import com.laytonsmith.abstraction.blocks.MCMaterial;
+import com.laytonsmith.abstraction.enums.MCVersion;
 import com.laytonsmith.annotations.api;
 import com.laytonsmith.annotations.breakable;
 import com.laytonsmith.annotations.core;
 import com.laytonsmith.annotations.seealso;
 import com.laytonsmith.core.ArgumentValidation;
 import com.laytonsmith.core.CHVersion;
-import com.laytonsmith.core.ObjectGenerator;
 import com.laytonsmith.core.Optimizable;
 import com.laytonsmith.core.ParseTree;
 import com.laytonsmith.core.Script;
@@ -1349,12 +1351,12 @@ public class BasicLogic {
 
 		@Override
 		public Class<? extends CREThrowable>[] thrown() {
-			return new Class[] {CREFormatException.class, CRENotFoundException.class};
+			return new Class[] {CREFormatException.class, CRENotFoundException.class, CRECastException.class};
 		}
 
 		@Override
 		public String docs() {
-			return "boolean {itemArray1, itemArray2} Returns true or false if the item are equal.";
+			return "boolean {itemArray1, itemArray2} Returns true if the items are equal.";
 		}
 
 		@Override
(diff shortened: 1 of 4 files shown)


PseudoKnight:
Well, an item array comparison would be faster, particularly for non-matches, and would give the script more control over what the scripter wants to compare. 

Reminds me. I wrote an array function that checks if one array is a subset of another array. I wonder if that would be useful to people. I was going to use it for an item array prefilter.

src/main/java/com/laytonsmith/core/functions/BasicLogic.java
```diff
@@ -1328,6 +1331,55 @@ public CBoolean exec(Target t, Environment environment, Construct... args) throw
...
+		}
+
+		@Override
+		public Class<? extends CREThrowable>[] thrown() {
+			return new Class[] {CREFormatException.class, CRENotFoundException.class};
+		}
+
+		@Override
+		public String docs() {
+			return "boolean {itemArray1, itemArray2} Returns true or false if the item are equal.";
```
LadyCailin: The phrasing on this should state: "Returns true if the items are equal."

EntryPointKR:
@PseudoKnight Thanks for your advice

I think If create string equals method in CArray class, duplicate val().equals() can be removed.
How do you think?

PseudoKnight:
@EntryPointKR String equals is not equivalent to isSimilar(). It'd be more strict, sometimes evaluating false in situations where the items would be the same. In a script, you'd just compare values in each key, while ignoring keys that don't apply. isSimilar() ignores quantity, for example. Comparing meta would be a little complex, but most items don't have meta and wouldn't even need to check it. I wrote the start of a similar script to give you an idea.

````
proc _is_similar(@a, @b) {
	if(@a['type'] != @b['type']) {
		return(false);
	}
	if(array_get(@a, 'data', 0) != array_get(@b, 'data', 0)) {
		return(false);
	}
	if(array_index_exists(@a, 'meta') && !is_null(@a['meta'])
	|| array_index_exists(@b, 'meta') && !is_null(@b['meta'])) {
		@metaA = array_get(@a, 'meta', associative_array());
		@metaB = array_get(@b, 'meta', associative_array());
		if(array_get(@metaA, 'display', '') != array_get(@metaB, 'display', '')) {
			return(false);
		}
		... etc
	}
	return(true);
}
````

This is useful even if equals_item() is added. It probably could be improved, though.

**Comment by EntryPointKR** [comment]
I added equals method in Construct class.
What do you think?

**Comment by PseudoKnight** [comment]
I think you misinterpreted what I was saying. I didn't even notice you changed the comparison method. I never meant for you to write the array comparison in the Java function. I'm sorry. My first response to new functions is "does this need to be a function?" I was starting a discussion about that.

**Comment by EntryPointKR** [comment]
I knew CommandHelper was a plugin that does not require Java knowledge for convenience. If I first set up the CommandHelper, I had to wrote a procedure which compares two items. I wanted to use this feature in core, just as I would use the isSimilar() method in Java without having to compare type, data and meta etc.. so I sent a pull request.

**Comment by EntryPointKR** [comment]
and I don't know why failed the test.

**Comment by PseudoKnight** [comment]
If this were added as a function, I'd probably favor a combination where it checked the type and data first, then maybe check if the meta even exists on either item. Only then convert the items to MCItemStack to use in isSimilar() for the meta comparison. That way we don't have to add complex meta checks and maintain them here too.

What do others think? Also, where would this function be used? I can't think of a scenario right now.

**Comment by EntryPointKR** [comment]
I believe, this function will be useful in all situations when comparing item.
For example
```
bind(player_interact, null, null, @e) {
    @handItem = pinv(player(), null);
    if (item_equals(@handItem, _get_usable_item())) {
        // ACTION
    }
}

/giveitem = >>>
    @item = _get_usable_item();
    pgive_item(player(), @item['type'].':'.@item['data'], 1, @item['meta']);
<<<

proc _get_usable_item() {
    return(array(
        'data': 0,
        'type': 280,
        'qty': 1,
        'meta': array(
            'display': 'Simply',
            'lore': 'abc'
        )
    ));
}
```
If we don't use item_equals() then I think this feature won't be efficiently used as much.
```
bind(player_interact, null, null, @e) {
    @handItem = pinv(player(), null);
    if (@handItem && @handItem['type'] == 280 && @handItem['data'] == 0 && meta) {
        // ACTION
    }
}
```

PseudoKnight:
It's not as efficiently written because in one you're using a proc and the other you're not. 

My subMap function would be handy in that scenario. Still not sure what to call it. But that was my attempt at a more elegant solution.
````
bind(player_interact, null, null, @e) {
    @handItem = pinv(player(), null);
    if (array_subset_of(_get_usable_item(), @handItem)) {
        // ACTION
    }
}
````

EntryPointKR:
I think your idea is right.
Thanks for your answer.
It was good experience for me

EntryPointKR:
@PseudoKnight 
Can you tell me about array_subset_of()?
I don't know exactly what the function is.

PseudoKnight:
I mentioned it earlier. "Reminds me. I wrote an array function that checks if one array is a subset of another array. I wonder if that would be useful to people. I was going to use it for an item array prefilter." With it you could see if {type: 1, meta: {display: Title}} was a subset of a typical ObjectGenerator item array output like {name: STONE, type: 1, qty: 1, data: 0, enchants: [], meta: {display: Title, lore: null, flags: [], enchants: []}}. It would return true in this case. However, if meta was null, as most STONE is, it would return false. I could imagine it being useful in other cases too (like location arrays), but the most useful case would be for item arrays, I think. It's more of LC's domain, though, so I wasn't confident enough to PR it.

EntryPointKR:
Is array_subset_of () a function that does not currently exist? If yes, can I pull request it? actually, I wrote this function before ask about array_subset_of(). but I wanted do it exactly.

PR #445, opened 2016-12-13
Add entity_regain_health event

EntryPointKR:
Is @EventIdentifier works for compatibility reason?


PseudoKnight:
Should add entity "id" to the event data. Mention "id" in the documentation as well as "player", which is added to entity events by evaluate_helper() when the entity type is a player. modifyEvent() should get a double, not an int, for "amount".

EntryPointKR:
Fixed. thanks
and I want to know about @EventIdentifier annotation.
Why use it?

PseudoKnight:
That was created by jbaero because different versions of the Bukkit API might not have all the events. So in events/Driver.java we can specify which version to attempt to register a corresponding listener, like with ENTITY_TOGGLE_GLIDE which doesn't exist prior to Minecraft 1.9.

EntryPointKR:
Always thanks for your answer and sorry to my bad English

PseudoKnight:
This seems good to go, @LadyCailin

PR #446, opened 2016-12-20
Force BukkitGetCorrectEntity to determine a more specific entity class for UNKNOWN entity type

PseudoKnight:
PRing this just to make sure I'm not missing something here. (@jb-aero) 

UNKNOWN entity types are sometimes handled as more specific classes than just MCEntity. For example, currently when EchoPet causes a CreatureSpawnEvent the entity type is UNKNOWN. CH wraps it as an MCEntity and causes an exception when BukkitMCCreatureSpawnEvent tries to cast it to MCLivingEntity. By setting the wrapperclass as null, it forces BukkitConvertor.BukkitGetCorrectEntity() to find the most specific entity class.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/enums/bukkit/BukkitMCEntityType.java
+++ b/src/main/java/com/laytonsmith/abstraction/enums/bukkit/BukkitMCEntityType.java
@@ -110,7 +110,7 @@ public class BukkitMCEntityType extends MCEntityType<EntityType> {
 	private void setWrapperClass() {
 		switch (getAbstracted()) {
 			case UNKNOWN:
-				wrapperClass = BukkitMCEntity.class;
+				wrapperClass = null;
 				break;
 			case DROPPED_ITEM:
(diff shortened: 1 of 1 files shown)


Borlea:
@PseudoKnight If you have any suggestions on a way to support such things on EchoPets i'll be glad to implement it in order to prevent any other issues with peoples plugins.

PseudoKnight:
@Borlea The only issue here is that it's an UNKNOWN entity type, which is unusual, and CH needs to handle them more specifically. Our newer ENUM converter (to handle multiple versions of Minecraft better, as more and more servers spread out between 1.7.10 and 1.11) just didn't fully take that edge case into consideration. I'm sure there's a reason for that entity type in EchoPet, given what it's doing. We can't expect compatibility with all plugins, let alone NMS plugins, but this looks like an easy fix that I don't think will have any repercussions. 

It seems @jb-aero is busy, so I'll just merge this.

PR #448, opened 2017-01-05
Removed 16 char limit from set_list_name().

Pieter12345:
The 16 character list name limit has been removed from Bukkit. The exact
limit is unknown and depends on the client packet receive code, but it's
at least 30.000.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
+++ b/src/main/java/com/laytonsmith/core/functions/PlayerManagement.java
@@ -3496,16 +3496,14 @@ public class PlayerManagement {
 
 		@Override
 		public Construct exec(Target t, Environment environment, Construct... args) throws ConfigRuntimeException {
-			MCPlayer m = environment.getEnv(CommandHelperEnvironment.class).GetPlayer();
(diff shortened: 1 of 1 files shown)


PseudoKnight:
This check was removed in 1.8. It'll cause an IllegalArgumentException on 1.7.10. This should probably be caught and rethrown as a CRELengthException until support is dropped for that version.

https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/commits/24557bc2b37deb6a0edf497d547471832457b1dd#src/main/java/org/bukkit/craftbukkit/entity/CraftPlayer.java

However, since documentation should at least match latest versions, the limit should be removed from the docs() and probably CRELengthException.class removed from thrown().

PseudoKnight:
Not sure if we should mention things like "Will throw a LengthException on 1.7.10 if the name exceeds 16 characters." We haven't done that in the past, but I don't CH has really supported this wide a range of versions before.

PseudoKnight:
Just remembered CRELengthException is thrown by Static.GetPlayer() anyway. 

Ya, this looks good to me.

PR #450, opened 2017-01-11
Reduce overhead when creating and modifying arrays

PseudoKnight:
This avoids hashing entire arrays while still preventing recursion. In most cases they weren't noticeable, but these hashes could add up when adding or setting many values in an array, even internally like with inventory events.

### DIFF
--- a/src/main/java/com/laytonsmith/core/constructs/CArray.java
+++ b/src/main/java/com/laytonsmith/core/constructs/CArray.java
@@ -119,7 +119,7 @@ public class CArray extends Construct implements ArrayAccess{
 			}
 			this.next_index = array.size();
 		}
-		regenValue(new HashSet<CArray>());
+		regenValue();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


PseudoKnight:
I think there's cases where this wouldn't work as expected. May need to go a different route.

PseudoKnight:
Also realized the previous logic didn't work in all cases either. Can't wait to get rid of auto concat in 2023.

PseudoKnight:
There. It now safely avoids hashing when unnecessary. I even found that it wasn't using cached strings for sub arrays, so I fixed that. I added a test that my previous implementation would have broken. Also, there's a commented out test for something that has never worked, but should be noted anyway. When an array has multiple parents and is modified, it will only set the dirty flag for the latest parent array. This wasn't in the scope of what I was trying to fix in this PR, so I left it.

PseudoKnight:
Improved this and pushed in https://github.com/EngineHub/CommandHelper/commit/5058dae47605a069cb4cd77f777c919044e76964

PR #455, opened 2017-02-03
Add get|set_world_border

PseudoKnight:
Chose "width" instead of "size", as it's more clear.


PseudoKnight:
Found a bug with more testing.

PseudoKnight:
Never mind. It was just worldborders themselves being weird.

PR #456, opened 2017-02-24
Add block_fade event

EntryPointKR:
Add block_fade event

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitBlockListener.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitBlockListener.java
@@ -124,4 +124,10 @@ public class BukkitBlockListener implements Listener{
 		BukkitBlockEvents.BukkitMCNotePlayEvent npe = new BukkitBlockEvents.BukkitMCNotePlayEvent(e);
 		EventUtils.TriggerListener(Driver.NOTE_PLAY, "note_play", npe);
 	}
+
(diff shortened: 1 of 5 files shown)


PseudoKnight:
Is there any case where data won't be 0?

src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitBlockListener.java
```diff
@@ -124,4 +124,10 @@ public void onNotePlay(NotePlayEvent e) {
 		BukkitBlockEvents.BukkitMCNotePlayEvent npe = new BukkitBlockEvents.BukkitMCNotePlayEvent(e);
 		EventUtils.TriggerListener(Driver.NOTE_PLAY, "note_play", npe);
 	}
+
+	@EventHandler(priority = EventPriority.LOWEST)
+	public void onBloackFade(BlockFadeEvent e) {
```
PseudoKnight: Bloack.

EntryPointKR:
"data" means e.getBlock.getData() ?

PseudoKnight:
What block can fade that has data other than 0?

EntryPointKR:
51(fire), others I don't know

PseudoKnight:
Oh, right. I can't think of a reason to prefilter that, though. I think olddata, newtype, and newdata should be removed from prefilters, as they serve no purpose. Only oldtype and world are meaningful prefilters here.

EntryPointKR:
Okay I'll remove it now

PR #457, opened 2017-06-08
Added useSSL to MySQLProfile

Ecconia:
An optional additional argument to enable/disable ssl.
Not adding the <useSSL> Tag, will default to not sending the argument.
<useSSL /> will default to false, any mix of true/TRUE will be interpreted as true.

### DIFF
--- a/src/main/java/com/laytonsmith/database/MySQLProfile.java
+++ b/src/main/java/com/laytonsmith/database/MySQLProfile.java
@@ -17,6 +17,7 @@ public class MySQLProfile extends SQLProfile {
 	private final String database;
 	private final String username;
 	private final String password;
+	private final Boolean useSSL;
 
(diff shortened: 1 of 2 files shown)


Ecconia:
Added since MySQL was bugging me to send it after each recompile:

"WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification."

The connection does not need to be secure since it stays on a local machine.

LadyCailin:
Looks good, thanks for the submission!

Ecconia:
Thanks for quick merging 👍

PR #458, opened 2017-07-29
Make uncaught exception handler plugin-wide, and pass commandsender…

PseudoKnight:
…from exception env into handler.

This is what Pieter was presumably trying to solve in PR #294, however it doesn't keep it separate from /interpreter as talked about. Right now I'm setting the exception handler in auto_include.ms to get around this problem, but I'm wondering if it really makes sense to pass it around in the environment like it is.

The second part to this PR (even if the first part isn't approved) is to send the commandsender from the exception's environment to the exception handler. This way we can send error messages to the user too. This is normally not a problem in aliases, but it is in registered commands and other places where the commandsender for when the exception handler is created is not the same as the commandsender for when the code is run. This change is far more important to me at the moment.

### DIFF
--- a/src/main/java/com/laytonsmith/core/environments/GlobalEnv.java
+++ b/src/main/java/com/laytonsmith/core/environments/GlobalEnv.java
@@ -8,7 +8,6 @@ import com.laytonsmith.core.Procedure;
 import com.laytonsmith.core.Profiles;
 import com.laytonsmith.core.Script;
 import com.laytonsmith.core.Static;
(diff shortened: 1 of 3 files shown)


PseudoKnight:
I should clarify that I'm not entirely clear what the scope of the uncaught exception handler is SUPPOSED to be. It does feel like an unusual case. Should it be passed around or should scripts be able to set it once in main.ms, for example. My original intention was to just fix the exception commandsender, but I thought I'd take a look at this "problem" too.

PR #462, opened 2017-09-29
Fix bukkit yaw in an earlier stage.

Pieter12345:
Bukkit can return any yaw value, while we except a yaw to be in range of
0 to 360. This commit makes sure that any BukkitMCLocations will return
that range, rather than (partially) fixing it in a later stage.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCLocation.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCLocation.java
@@ -68,7 +68,15 @@ public class BukkitMCLocation implements MCLocation {
 
 	@Override
     public float getYaw() {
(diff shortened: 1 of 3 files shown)


PseudoKnight:
I know this was my thought, but I think I'll keep the 0-360 guarantee in construct creation.

Pieter12345:
@PseudoKnight In that case, also add it to ObjectGenerator.location(Location, boolean), since it's not there currently. The ObjectGenerator.location(Location) method code can still be replaced with a call to ObjectGenerator.location(Location, true).

Pieter12345:
@PseudoKnight And in pfacing(), the yaw is modulo'd to the proper range, effectively casting it to an integer instead of a float/double. This should also get an appropriate fix then.

PR #463, opened 2017-09-29
Add title()

PseudoKnight:
Last chance to comment on the name, parameters and documentation.

Other proposed names included send_title(), psend_title(), show_title(), and title_msg(). These were for the most part not any more descriptive, so I decided to keep it simple, just like most other echo functions.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCPlayer.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCPlayer.java
@@ -341,6 +341,20 @@ public class BukkitMCPlayer extends BukkitMCHumanEntity implements MCPlayer, MCC
     }
 
 	@Override
+	public void sendTitle(String title, String subtitle, int fadein, int stay, int fadeout) {
+		try {
+			p.sendTitle(title, subtitle, fadein, stay, fadeout);
+		} catch(NoSuchMethodError ex1){
+			// Probably prior to 1.11, try the deprecated method
+			try {
+				p.sendTitle(title, subtitle);
+			} catch(NoSuchMethodError ex2){
+				// Probably prior to 1.8.7, no title API
+			}
+		}
+	}
+
+	@Override
     public void setAllowFlight(boolean flight) {
         p.setAllowFlight(flight);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCPlayer.java
```diff
@@ -341,6 +341,20 @@ public void sendResourcePack(String url) {
...
 	@Override
+	public void sendTitle(String title, String subtitle, int fadein, int stay, int fadeout) {
+		try {
+			p.sendTitle(title, subtitle, fadein, stay, fadeout);
+		} catch(NoSuchMethodError ex1){
+			// Probably prior to 1.11, try the deprecated method
+			try {
+				p.sendTitle(title, subtitle);
+			} catch(NoSuchMethodError ex2){
+				// Probably prior to 1.8.7, no title API
```
Pieter12345: I would throw some CRE here so that the function doesn't fail seemlessly. This allows a methodscript programmer to detect that this function does not work and give him/her the choice to update their server software or don't use this function.

PseudoKnight: Currently every method fails silently if possible (or returns a neutral value) when it's not supported by their version of Bukkit, as long as it's 1.7.10 or later. Server admins usually make a conscious decision to stay on older versions in lieu of new features. One reason is I don't want to break shared scripts by interrupting them for something that doesn't change state. CREs usually aren't generated in Abstraction and I like to avoid adding multi-version support in the function code instead of the Abstraction. Does that make sense or do I need to change all of those methods?

PseudoKnight: I'd like to talk sometime about guidelines on how to deal with this issue, but for this commit I'll leave it like the others.

Pieter12345: I get your point, and it is indeed the question whether we want something to happen for unsupported Bukkit versions. I gave it some thought and I think a good way to handle these cases would be to log a warning (+ console message) based on some boolean in the config. The abstraction method would return without an Exception, and would call a method which logs the problem. An alternative is to throw an Exception anyways, but based on some boolean in the config (don't throw if that boolean has some value).

PR #464, opened 2017-10-17
Optimize alias permission labels

PseudoKnight:
Script labels are now processed only once, instead of every time a function is executed. This also isolates script labels so that they're not modified after they're set, which is what the environment is for. These changes should not alter current behavior.

I don't expect any issues, but I want to do more testing before this is pulled. However, I want your impression on this approach and if you have any suggestions. I don't want to do the testing if you don't like it, and I had trouble communicating what I wanted to do here before.

So the git diff on the scriptas() function turned out pretty shitty. What I did there was removed script.setLabel() usages (as we probably shouldn't ever change the script label) and added the same script.enforceLabelPermissions() call that's at the beginning of script.run().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,10 +145,6 @@
 			<id>maven-central</id>
 			<url>http://repo1.maven.org/maven2/</url>
 		</repository>
-		<!--repository>
-			<id>zml2008</id>
-			<url>http://files.zachsthings.com/repo/</url>
-		</repository-->
 
 		<repository>
 			<id>Plugin Metrics</id>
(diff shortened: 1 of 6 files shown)


PseudoKnight:
Oh ya, this also fixed a bug in the initial group label check which would have caused an insufficient permission exception with a particular unsupported usage. (eg. "-label:/cmd")

LadyCailin:
This looks fine! This doesn't change the behavior at all, this is purely optimization, so I'm totally ok with this. I would want to have a bigger discussion about the individual function permissions, but that's a separate issue anyways. Looks good!

PR #465, opened 2017-10-25
Change Minecraft reflection method to improve startup times

PseudoKnight:
Despite only being used for sudo() right now, ClassDiscovery for the server jar adds up to 10 seconds to the startup time. The Bukkit convertor init also synchronizes during the ClassDiscovery process which may make things worse. This commit proposes that this extra time is not a necessary sacrifice and implements an alternative reflection method where needed.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCPlayer.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCPlayer.java
@@ -2,7 +2,6 @@
 
 package com.laytonsmith.abstraction.bukkit.entities;
 
-import com.laytonsmith.PureUtilities.ClassLoading.ClassDiscovery;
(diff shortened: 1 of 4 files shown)


PseudoKnight:
I should explain the symptoms. When starting up it freezes for around 10 seconds on StaticLayer.GetServer(), presumably because the StaticLayer.InitConvertor() is waiting on the ClassDiscovery to finish. After this commit, it now freezes only for a second or two after "Waiting for extension loading to complete...". Overall this commit cuts my test server's startup time by over half. (from 24-31 seconds to 10-11 seconds).

LadyCailin:
This is ok. My only concern is, will the first call to sudo block now?

PseudoKnight:
If you mean that it'd wait for the convertor init, that's just called later now onEnable(). If you mean it'd be slower, it didn't appear to be slower. It's simply not using the ClassDiscovery system now, instead relying on  basic reflection.

PR #466, opened 2017-10-31
Added missing Prefix

Ecconia:
The prefix in this feedback was missing. The logger does not add it by default here.

### DIFF
--- a/src/main/java/com/laytonsmith/PureUtilities/ClassLoading/ClassDiscoveryCache.java
+++ b/src/main/java/com/laytonsmith/PureUtilities/ClassLoading/ClassDiscoveryCache.java
@@ -130,7 +130,7 @@ public class ClassDiscoveryCache {
 			}
 			
 			if(logger != null){
-				logger.log(Level.INFO, "Performing one time scan of {0}, this may take a few moments.", fromClassLocation);
+				logger.log(Level.INFO, "[CommandHelper] Performing one time scan of {0}, this may take a few moments.", fromClassLocation);
 			}
 			
 			ClassDiscoveryURLCache cache = new ClassDiscoveryURLCache(fromClassLocation, progress);


PseudoKnight:
Is it missing? It shows up on my test server. Maybe it's being truncated by your console viewer?

[15:10:14] [Server thread/INFO]: [ProtocolLib] Loading ProtocolLib v4.4.0-SNAPSHOT-b390
[15:10:14] [extensionloader/INFO]: [CommandHelper] Performing one time scan of file:/E:/Desktop/Servers/1.12.1/plugins/CommandHelper/extensions/CHDynmap-1.1.3.jar, this may take a few moments.
[15:10:15] [extensionloader/INFO]: [CommandHelper] Performing one time scan of file:/E:/Desktop/Servers/1.12.1/plugins/CommandHelper/extensions/CHFiles-2.1.3.jar, this may take a few moments.
[15:10:15] [Server thread/INFO]: [WorldGuard] Loading WorldGuard v6.1.3-SNAPSHOT;324e90c

It looks like it's using Logger.getLogger(CommandHelperPlugin.class.getName()) in onLoad().

PseudoKnight:
Oh, on Paper it prints the Logger label. Spigot does not. So not sure of the best course of action. We could stop using Logger, I guess, or we could withstand the double printing "[CommandHelper]".

[15:10:27] [Server thread/INFO]: [CommandHelper] [CommandHelper]: MethodScript files processed

Ecconia:
I was running it with CraftBukkit 1.12.2 maybe also worth knowing. 

Regards the branding, okay didn't know that, I just followed a code example related to this code, which also uses hard coded prefixes. 

I tried both (already used) ways of adding a logger (Static.getLogger and Logger.getName) both didn't print the prefix for me. 
Let me quickly check if that was also the case for getLogger from the JavaPlugin class, that one has a prefix for sure. But I don't know if that is still present if passed down to that Thread.

PseudoKnight:
Ya, I'm thinking either we need to unify under CHLog or Static.getLogger(), then we can set those to use the plugin logger if the implementation is BUKKIT.

Ecconia:
Yes, on CraftBukkit getLogger() will print the prefix. I tried to figure out, why Logger.getName does not resolve to that Logger. Sadly I did not find any reason or solution.
Probably saving and using the plugin logger would be a solution.

PseudoKnight:
CHLog looks like it's more designed for warnings, errors, and debug. So either that needs to be changed or we should probably use Static.getLogger() when we always want to print and have Static keep the appropriate logger. Either way Static should probably have the appropriate logger because CHLog uses it too.

I know LadyCailin talked about redoing logging, but she never mentioned what she had planned, just that it was poor currently.

Ecconia:
It seems that my attempt on how to solve this issue was plainly wrong. 
A logger should handle the prefix itself. 

You may close this PR, since it has to be fixed in a different way and in a larger scale.

PseudoKnight:
http://dev.enginehub.org/youtrack/issue/CMDHELPER-3163

PR #471, opened 2018-01-30
Improved runtime complexity GetConstruct/GetPOJO

Pieter12345:
Both the GetConstruct and GetPOJO methods contained loops over
Lists/Maps with a getter of a value from those Lists/Maps in them,
effectively being O(n^2) runtime complexity (with an exception of Hash
implementations, where a lookup is O(1) instead of O(n)).
In this commit, those loops have been changed to loops over the entryset
(for Maps) or changed to using an iterator (Lists). The new runtime
complexity for both is O(n).


cailin-telenor:
👍

PR #472, opened 2018-02-03
Compatibility fixes

Pieter12345:
- Added <pluginManagement> tag so that Eclipse's integrated Maven module can handle the <execute> tag.
- Added URL decoding and encoding to support having a java eclipse (and probably other IDE's too) workspace with whitespaces in the path (before this commit, many tests would be unable to run due to ClassInitializationErrors).


dermarionator:
Is http://www.sk89q.com down?
I can't reach the server. Error 522

Pieter12345:
@dermarionator This is not the place to ask random questions. You can ask questions like these on IRC (web client https://esper.net/publicirc.php, channel #CommandHelper or #sk89q).

Pieter12345:
Removed pluginManagement tag from the pom since it did fix the error, but still didn't execute some plugins on Eclipse (so essentially didn't fix anything). The remaining changes fix a bug in CommandHelper that happens when running tests while having a whitespace in the workspace (due to whitespaces in URL's being "%20", which does not lead to the actual file when used).

PR #475, opened 2018-02-14
BugFix for ambigous command detection.

Pieter12345:
Many commands were not detected as ambigous properly. Examples:
- *:/cmd2 $arg AND *:/cmd2 $arg
- *:/cmd11 $arg $arg2 AND *:/cmd11 $arg $arg2 [$arg3]
- *:/cmd14 arg AND *:/cmd14 $arg
- *:/cmd15 arg AND *:/cmd15 [$arg]

This commit will cause ambigous commands to throw a compile exception as
expected.
The previous code caught an IndexOutOfBoundsException, which's
fillInStacktrace() costed ~35% runtime spent in
AliasCore.LocalPackage.consoleMSA(). This has been replaced with a size
check to reduce runtime for MSA compiles by roughly this 35% (as tested
with my scripts and setup on my server with 254 defined aliases).

### DIFF
--- a/src/main/java/com/laytonsmith/core/Script.java
+++ b/src/main/java/com/laytonsmith/core/Script.java
@@ -841,10 +841,10 @@ public class Script {
 					if(k < thatCommand.size()) {
 						Construct c2 = thatCommand.get(k);
 						
-						// Commands are not ambigous if they have unequal commands or literals at
+						// Commands are not ambigous if they have unequal commands or strings at
 						// the same argument position.
 						if(c1.getCType() == c2.getCType()
-								&& (c1.getCType() == ConstructType.LITERAL || c1.getCType() == ConstructType.COMMAND)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/laytonsmith/core/Script.java
```diff
@@ -486,7 +484,7 @@ public boolean match(String command) {
 		}
 		boolean case_sensitive = Prefs.CaseSensitive();
 		String[] cmds = command.split(" ");
-		List<String> args = new ArrayList(Arrays.asList(cmds));
+		List<String> args = new ArrayList<String>(Arrays.asList(cmds));
```
LadyCailin: new ArrayList<>

LadyCailin:
Would be nice to have some unit tests on this. They probably would have revealed the problem long ago had I done that in the first place.

src/main/java/com/laytonsmith/core/Script.java
Pieter12345: Changed.

PR #481, opened 2018-04-28
Added broadcast_message event.

Pieter12345:
Added support for Bukkit's BroadcastMessageEvent.
The recipients of the Bukkit event cannot be changed and only the player recipients are passed to CH. The other CommandSenders can be console, normal command blocks, command block functions and probably more depending on the server software used. Except for the console sender, these other non-player senders do not have a unique name and/or location and seem therefore useless to include.


PseudoKnight:
This was added in 1.12 and should be dynamically registered. I could do that for you, if you want.

Pieter12345:
@PseudoKnight Sure, I've never done that with events, so go ahead :)

PR #482, opened 2018-04-29
fix pre browser inheritance behavior

good3n:
Per Pieter12345's suggestion, text will wrap when necessary and/or on line-breaks.

### DIFF
--- a/src/main/resources/siteDeploy/resources/css/styles.css
+++ b/src/main/resources/siteDeploy/resources/css/styles.css
@@ -27,7 +27,8 @@ table td p {
     border-top: 1px solid #ccc;
     border-bottom: 1px solid #ccc;
     box-shadow: 0 .3em .3em rgba(0, 0, 0, 0.1);
-    color: #555
+    color: #555;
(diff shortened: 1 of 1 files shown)


Pieter12345:
To illustrate what this does:
- Before: https://imgur.com/iu1TYRf
- After: https://imgur.com/68pdyfv

LadyCailin:
Closed?

PseudoKnight:
We were discussing what would be the most appropriate change. Something needs to be done, but there are a few options. I wasn't sure what you'd prefer.

"white-space: pre-line" is good if you want all the information available inside the box and ignore white-space.

"white-space: pre-wrap" also wraps but doesn't collapse white-space.

"overflow: auto" would just add a horizontal scroll bar if we want to perfectly respect existing new lines. (no wrapping) This would be best for code, but we're using a div for that.

PseudoKnight:
Looking at some uses of pre, seems we need to preserve whitespace and not collapse it. (eg. Persistence Network benchmark table) So pre-wrap is probably want we want.

PR #485, opened 2018-05-22
CheckStyle: Add more rules [Do not merge before checkstyle violation fix commits]

Pieter12345:
A CheckStyle configuration which will ensure more consistent sourcecode in the future.


Pieter12345:
@LadyCailin Feel free to leave questions and comments, I will be away for about a week.

PseudoKnight:
I used fallthrough intentionally in at least one location, and there may be others.

Some of this seems a little aggressive, considering the primary reason we wanted style checks was for tab consistency. The MissingCtor one would require us to change every function, for example. Is this busy work?

Pieter12345:
@PseudoKnight Is this fallthrough from a non-empty code block? Otherwise it'd be allowed. The MissingCtor could be disabled for the com.laytonsmith.core.function package, like the ClassNameFormat.

Regarding aggressiveness, I've added checks I thought might be good for CommandHelper with the intention to get more consistent code without forcing contributors to change their method of coding (it shouldn't get in the way). If checks appear to have violations where a violation fix in the code seems unnecessary, the check should be changed or removed instead. Once I'll start fixing violations, I'll automatically run into these. So as for this PR, it's just here so LadyCailin, you can other contributors can view the rules I intend to add before I actually start "fixing" them.

PseudoKnight:
Yes, the fallthrough is from an non-empty code block, though I thought that was implied.

When evaluating these, just remember that it should make coding contributions easier, not harder.

checkstyle.xml
```diff
@@ -2,47 +2,219 @@
...
+		<module name="MissingDeprecated"/>
+		
+		<!-- Ensure that the @Override tag is present when the {@inheritDoc} javadoc tag is present -->
+		<module name="MissingOverride"/>
+		
+		<!-- Ensure modifier order is 'first > last': public > protected > private > abstract > default > static > final > transient > volatile > synchronized > native > strictfp -->
+		<module name="ModifierOrder"/>
+		
+		<!-- Ensure that each variable has its own statement on its own line (disallows "i = j = k = 1;") -->
+		<module name="ModifierOrder"/>
```
LadyCailin: You have ModifierOrder twice here? Certainly that is not correct.

checkstyle.xml
```diff
@@ -2,47 +2,219 @@
...
+		<module name="SeparatorWrap">
+			<property name="option" value="eol"/>
+			<property name="tokens" value="COMMA"/>
+		</module>
+		
+		<!-- Ensure non-static, non-final field name format. Default format: "^[a-z][a-zA-Z0-9]*$" -->
+		<module name="StaticVariableName"/>
+		
+		<!-- Disallow the "==" operator for string literals -->
+		<module name="StringLiteralEquality"/>
```
LadyCailin: There may be a few places where this is a problem, because I am validly using == in place of .equals, so I'm curious to see the results of this. Certainly though, in 90% of the cases, this will be an error in the code, so perhaps we just do //CHECKSTYLE:OFF on those blocks.

checkstyle.xml
Pieter12345: This is indeed not correct, the second one should have been "MultipleVariableDeclarations".

checkstyle.xml
Pieter12345: This should only fire when a string literal (and not a variable) is checked using "==". It is still possible to perform `someVar == someOtherVar`, but not `someVar == "str"` (which would always be false anyways).

Pieter12345: @LadyCailin This rule currently has one violation in `BukkitBlockEvents:447`, for code : `item.getType().getName() == "AIR"`, which is luckily only used as a condition to throw a CRE in a very specific situation at the moment, but it definately is a mistake which could lead to a failure/bug.

src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCWorld.java
```diff
@@ -640,6 +640,7 @@ public CArray spawnMob(MCMobs name, String subClass, int qty, MCLocation l, Targ
 								switch(ztype) {
 									case HUSK:
 										mobType = Husk.class;
+										continue;
```
PseudoKnight: This is incorrect.

PseudoKnight: I know it's weird, shush! This is an example of the stupid stuff I had to for 1.11 to avoid making an even more complicated backwards compatibility system for a single function that doesn't even use the proper entity ENUM.

cailin-telenor:
After giving this more thought, I think we should add the no-fallthrough thing back. It can still be disabled manually with //CHECKSTYLE:OFF SwitchFallthrough. See this SO post for more details on how to implement this: https://stackoverflow.com/questions/4023185/how-to-disable-a-particular-checkstyle-rule-for-a-particular-line-of-code?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa Then in the cases where this is fully intentional, it can be disabled, but for the average case where this is most likely a mistake, we can still catch that.

LadyCailin:
Ooh, no, do it with annotations, not with comments.

@SupressWarnings("checkstyle:fallthrough")

src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCWorld.java
PseudoKnight: I was commenting on the wrong thing. This change is perfectly fine. Look before you leap.

PseudoKnight:
Annotations sound good. Apparently I misremembered where I used a non-empty fallthrough (maybe I changed it), so it wouldn't even be necessary there.

Pieter12345:
@LadyCailin I've re-added the FallThrough rule and suppressed its only remaining violation.

cailin-telenor:
Looks good! Just need to fix the errors I guess.

Pieter12345:
@LadyCailin At this point with this rule set, 26 violations are remaining. These are all name violations of public accessible fields/classes/annotations which I believe can break extensions when changed. I've suppressed the method name check errors for the same reason. Do you want to suppress these remaining violations, deprecate them and add an alternative or do something else with them?

LadyCailin:
@Pieter12345 In general, I like static methods to start with capital letters, but you're right, since there is a lot of public methods like that, we unfortunately cannot change that at this point. So let's just disable checkstyle for those and keep the rule going forward, if there's only 26 violations.

Pieter12345:
@LadyCailin It's time for magic, all violations have been resolved.

PR #486, opened 2018-06-08
JoinEvent inject Entity

Ecconia:
The function "set_metadata" and similar require an entity to be online. 
The entity (player) is known by the event already, but not the rest of the server. 
Injecting it, to make it accessible.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
@@ -817,6 +817,24 @@ public class PlayerEvents {
 			return e;
 		}
 
+		@Override
+		public void preExecution(Environment env, ActiveEvent activeEvent) {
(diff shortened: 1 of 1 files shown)


PseudoKnight:
Modifying the player entity is usually what player_join is for. Whereas player_login is mostly used to determine if the player should be allowed on the server or not. While the player entity object does exist at this time, the data in it is almost all defaults and much of it will even be overwritten. So when it throws PlayerOfflineException, this is strictly true, and it seems it could be more harmful than helpful to override this. So I'd be curious what use-case would benefit from this.

Ecconia:
Fail, im sorry. I changed the code in player_join and messed up the title.
The only hint to see that is the cast in the two methods. 
I will change the text, since it's very confusing.

PseudoKnight:
Oh, I see. I should have looked closer. This looks fine. Just in case, I looked over where this event is fired and while there might be a few idiosyncrasies, it's nothing that wouldn't also be experienced with player functions.

PR #490, opened 2018-06-21
Travis-CI test [DO NOT MERGE]

Pieter12345:
This PR is created to be able to trigger Travis-CI builds and time some code with the goal to detect why RandomTests.java generally takes 3 to >10 minutes (timeout) to complete.

### DIFF
--- a/src/test/java/com/laytonsmith/testing/RandomTests.java
+++ b/src/test/java/com/laytonsmith/testing/RandomTests.java
@@ -51,6 +51,7 @@ import java.lang.reflect.Method;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


LadyCailin:
Sure. Also get locales can safely be skipped.

Pieter12345:
The secure_string and get_locales boilerplate tests have been removed in https://github.com/EngineHub/CommandHelper/pull/491.
Travis-CI took 4 seconds to run RandomTests after the PR, where it sometimes timed out before (>10 minutes).
RandomTests is with 4 seconds still the slowest test in CH, which is acceptable in my opinion, so no more Travis CI testing has to be done. Therefore, this PR is no longer necessary.

PR #492, opened 2018-06-27
Fix broadcast() function event consistency

Pieter12345:
- Make broadcast(message, players) fire an event through its Bukkit implementation so that it matches the broadcast(message) and broadcast(message, permission) behaviour.
- Extend broadcast() documentation to include all behaviour.


PseudoKnight:
Maybe broadcast() should always include ~console?

PR #494, opened 2018-07-06
Add new inventory functions and virtual inventories.

PseudoKnight:
New functions: popen_inventory(), pviewing(), get_inventory_viewers(), get_virtual_inventories(), create_virtual_inventory(), and delete_virtual_inventory().
New "virtual" prefilter for inventory_click, inventory_drag, inventory_open, and inventory_close events.
New inventory "holder" data for inventory_open and inventory_close events.

So this is my attempt to bring virtual inventories to CommandHelper proper instead of having to rely on CHVirtualChests (and without breaking the extension). This allows us to use the same inventory functions and events for virtual inventories as we do for in-world inventories held by entities and blocks. I also used this opportunity to add some other useful functions: popen_inventory(), pviewing() and get_inventory_viewers(). This addresses a few tickets on the issue tracker. As always, I'm not entirely confident on the implementation and naming, so I'd love some feedback. 

I stuck with the "virtual" naming scheme used in CHVirtualChests since it's familiar to CommandHelper scripters, but deviated in other ways. For example, virtual inventories are no longer represented by a one dimensional array, but are instead created by a few function parameters. This makes persistence storing a little more verbose, but makes the overall functionality simpler and consistent.

Here's an example script that opens up a simple temporary inventory menu that runs the name of the item as a command:

```
@menu = array(
	1: array('name': 'STONE', 'meta': array('display': '/warp x')),
	2: array('name': 'SAND', 'meta': array('display': '/warp y')),
	3: array('name': 'DIRT', 'meta': array('display': '/warp z')),
);
create_virtual_inventory(player().'_warp_menu', 'HOPPER', 'Select a Warp:', @menu);
popen_inventory(player().'_warp_menu');

@clickBind = bind('inventory_click', null, array('virtual': true, 'player': player()), @event) {
	if(pviewing() == player().'_warp_menu') {
		cancel();
		run(@event['slotitem']['meta']['display']);
		play_sound(ploc(), array('sound': 'UI_BUTTON_CLICK'), player());
		close_pinv();
	}
}

bind('inventory_close', null, array('virtual': true), @event, @clickBind, @player = player()) {
	if(@player == player()) {
		unbind();
		unbind(@clickBind);
		delete_virtual_inventory(player().'_warp_menu');
		console('Debug: '.player().' closed '.@event['holder'], false);
	}
}
```
Here's some potential talking points:
- pviewing() returns null when it's a virtual inventory from another plugin (or if we were to create a virtual inventory without an id).
- You can't name a virtual id in the same format as an entity UUID since they're both strings when fetching an inventory by construct specifier.
- Unlike CHVirtualChests, when you create an inventory by the same id, it throws an exception instead of replacing it.
- The function and event key names are still up to debate.
- I could add an optional inventory array to popen_inventory(), but it wouldn't have an id, so you couldn't differentiate it from other virtual menus from other plugins.
- I filtered inventory types by those that can be opened and modified when virtual. However, I left redundant inventory types (eg. shulker box, which is essentially a chest). I could also filter those to make the list simpler.

Have any thoughts on these or anything else?

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/Convertor.java
+++ b/src/main/java/com/laytonsmith/abstraction/Convertor.java
@@ -128,6 +128,8 @@ public interface Convertor {
 	 */
 	MCInventory GetLocationInventory(MCLocation location);
 
+	MCInventoryHolder CreateInventoryHolder(String id);
+
 	/**
 	 * Run whenever the server is shutting down (or restarting). There is no guarantee provided as to what thread the
 	 * runnables actually run on, so you should ensure that the runnable executes it's actions on the appropriate thread
--- a/src/main/java/com/laytonsmith/abstraction/MCDoubleChest.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCDoubleChest.java
@@ -1,0 +1,5 @@
+package com.laytonsmith.abstraction;
+
+public interface MCDoubleChest extends MCInventoryHolder {
+	MCLocation getLocation();
+}
--- a/src/main/java/com/laytonsmith/abstraction/MCServer.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCServer.java
@@ -56,11 +56,9 @@ public interface MCServer extends AbstractionObject {
 
 	MCCommandMap getCommandMap();
 
-	MCInventory createInventory(MCInventoryHolder owner, MCInventoryType type);
+	MCInventory createInventory(MCInventoryHolder owner, MCInventoryType type, String title);
 
 	MCInventory createInventory(MCInventoryHolder owner, int size, String title);
-
-	MCInventory createInventory(MCInventoryHolder owner, int size);
 
 	/**
 	 * Provides access to local user data associated with a name. Depending on the implementation, a web lookup with the
--- a/src/main/java/com/laytonsmith/abstraction/MCVirtualInventoryHolder.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCVirtualInventoryHolder.java
@@ -1,0 +1,5 @@
+package com.laytonsmith.abstraction;
+
+public interface MCVirtualInventoryHolder extends MCInventoryHolder {
+	String getID();
+}
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitConvertor.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitConvertor.java
@@ -12,6 +12,7 @@ import com.laytonsmith.abstraction.MCEnchantment;
 import com.laytonsmith.abstraction.MCEntity;
 import com.laytonsmith.abstraction.MCFireworkBuilder;
 import com.laytonsmith.abstraction.MCInventory;
+import com.laytonsmith.abstraction.MCInventoryHolder;
 import com.laytonsmith.abstraction.MCItemMeta;
 import com.laytonsmith.abstraction.MCItemStack;
 import com.laytonsmith.abstraction.MCLocation;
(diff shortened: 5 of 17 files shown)


PseudoKnight:
Perhaps a good question is whether any of the functionality of functions or keys weren't immediately obvious by their names.

I just remembered another quirk. When doing pviewing() and the player is looking at their own inventory or no inventory at all (no way to differentiate between the two), it returns their own UUID. This feels odd, but there's no way around it. I noted it in the docs.

Pieter12345:
@PseudoKnight From Bukkit's JavaDocs:
`Gets the inventory view the player is currently viewing. If they do not have an inventory window open, it returns their internal crafting view.`
From CraftBukkit source, I can see that this `internal crafting view` is the `CraftInventoryCrafting` class in Bukkit. If this `CraftInventoryCrafting` is the 2x2 crafting inventory in the players GUI, it could be considered as never watched and ignored. The players inventory should be instance of some other class, allowing you to detect that as a valid inventory and returning null for the internal crafting view.

PseudoKnight:
What I was saying is that there's no way to differentiate whether the player is looking at their own inventory/crafting or nothing at all. If a player doesn't have an inventory open, it still returns the crafting view. So we'd have to to return null for both situations, if it makes sense to do so. The only reason I hesitated to do so is because it reduces the granularity of information the function provides. Returning null for that means we can't return null for other situations. Though admittedly it's weird using pviewing() == puuid() as means to check if a player is NOT viewing another inventory.

I've made changes based on your other suggestions and it's given me some things to think about. I'm still not sure I like pviewing_inventory(). The grammar is a bit odd. But while inventory_pviewing() is more clear, it might be too dissimilar to other function names.

PseudoKnight:
get_open_inventory() was one of the names I played around with early on.

PseudoKnight:
What about pinventory_holder()? It can't have "inventory" as the primary noun, as that's misleading (it doesn't return an inventory array). It can't have an ambiguous verb without "get_", which likely would make it the longest inventory function. (eg. get_popen_inventory_holder()) Could use something other than "holder", maybe even a generic identifier term (eg. id/specifier/target), but nothing strikes me better than holder.

EntityReborn:
I love this idea. xD

PseudoKnight:
Another possibility is using a past tense verb, popened_inventory_holder(), but I'm not too keen on that. I think the best we can do is make sure that the function reads intuitively in the contexts we're most likely to see it. I think we're most likely to see it used in an inventory event, but could also see it used to check if a player has an inventory open at a specific time. (I used the equivalent in chvirtualchests to check every second in a countdown if someone is still choosing a pvp class from an inventory menu) This is why pviewing() still worked for me despite its ambiguity. Examples:
```
bind('inventory_click', null, null, @event) {
    if(pviewing() == 'menu_id') {}
}

set_interval(1000, closure(){
    if(pviewing(@player) == 'menu_id') {}
});
```
Or...
```
bind('inventory_click', null, null, @event) {
    if(get_pview_inventory_holder() == 'menu_id') {}
}

set_interval(1000, closure(){
    if(get_pview_inventory_holder(@player) == 'menu_id') {}
});
```
Which also makes my last suggestion not as great:
```
bind('inventory_click', null, null, @event) {
    if(pinventory_holder() == 'menu_id') {}
}

set_interval(1000, closure(){
    if(pinventory_holder(@player) == 'menu_id') {}
});
```
I have half a mind to leave this function out for now and just use the "holder" key in inventory events. lol

PseudoKnight:
I also explored the shorthand "invholder", like with pinv_holder(), pviewing_invholder(), or get_popen_invholder().

PseudoKnight:
I'm in function naming hell. I think I'll just comment it out for now. In my first example I can just get the holder from the event.
```
bind('inventory_click', null, null, @event) {
    if(@event['holder'] == 'menu_id') {}
}
```
In my second example I can just see if the inventory viewers includes a player.
```
set_interval(1000, closure(){
    if(array_size(get_inventory_viewers('menu_id')) > 0) {}
});
```
There might still be use cases not yet served, but they're considerably rarer.

PseudoKnight:
I keep giving this more thought, thinking about what you all said and EntityReborn's positive reception to that one name. So, I changed the function to pinventory_holder() and no longer have virtual or holder keys in the inventory_click/drag events. I don't want to add even more overhead to that event, and it's not good to remove keys later.

It might be good to go if everyone is okay with that. I just now merged all the commits for one last overview.

PR #498, opened 2018-07-24
Fix set_uncaught_exception_handler closure return behaviour

Pieter12345:
Before this PR, the return value of the closure passed to set_uncaught_exception_handler would result in:
- true -> Suppress exception.
- null -> Use default uncaught exception handler.
- false -> Use default uncaught exception handler and print the java stacktrace to the cause.
- exception in closure -> Use default uncaught exception handler for the exception in the closure and suppress the first exception.

The behaviour set by this commit is:
- true -> Suppress exception.
- null -> Use default uncaught exception handler.
- false -> Use default uncaught exception handler.
- exception in closure -> Use default uncaught exception handler to print both exceptions.

This new behaviour matches the documentation of set_uncaught_exception_handler.

### DIFF
--- a/src/main/java/com/laytonsmith/core/exceptions/ConfigRuntimeException.java
+++ b/src/main/java/com/laytonsmith/core/exceptions/ConfigRuntimeException.java
@@ -100,10 +100,11 @@ public class ConfigRuntimeException extends RuntimeException {
 				return Reaction.REPORT; // Closure returned nothing -> REPORT.
 			} catch (FunctionReturnException retException) {
 				Construct ret = retException.getReturn();
-				if(ret instanceof CNull || Prefs.ScreamErrors() || !Static.getBoolean(ret, Target.UNKNOWN)) {
-					return Reaction.REPORT; // Closure returned null or false or scream-errors was set in the config.
+				if(ret instanceof CNull || Prefs.ScreamErrors()) {
+					return Reaction.REPORT; // Closure returned null or scream-errors was set in the config.
 				} else {
-					return Reaction.IGNORE; // Closure returned true -> IGNORE.
+					// Closure returned a boolean. TRUE -> IGNORE and FALSE -> FATAL.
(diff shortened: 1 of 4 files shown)


src/main/java/com/laytonsmith/core/exceptions/ConfigRuntimeException.java
```diff
@@ -86,32 +86,34 @@ public Environment getEnv() {
...
-				reaction = Reaction.REPORT;
-			} else {
-				if(Static.getBoolean(ret, Target.UNKNOWN)) {
-					reaction = Reaction.IGNORE;
+				Construct ret = retException.getReturn();
+				if(ret instanceof CNull || Prefs.ScreamErrors() || !Static.getBoolean(ret, Target.UNKNOWN)) {
+					return Reaction.REPORT; // Closure returned null or false or scream-errors was set in the config.
 				} else {
-					reaction = Reaction.FATAL;
+					return Reaction.IGNORE; // Closure returned true -> IGNORE.
```
LadyCailin: Hmm, no, we should still set this to FATAL, but wherever it is we handle it, it should be the same behavior

src/main/java/com/laytonsmith/tools/Interpreter.java
```diff
@@ -850,7 +850,8 @@ public static void install() {
 		if(null == OSUtils.GetOS()) {
 			StreamUtils.GetSystemErr().println("Cmdline MethodScript is only supported on Unix and Windows");
 			return;
-		} else switch(OSUtils.GetOS()) {
+		}
+		switch(OSUtils.GetOS()) {
```
LadyCailin: This happens to be ok in this case, but one cannot do this in general, as this is slightly different behavior than how it was.

Pieter12345: @LadyCailin Can you explain the different behavior? I would think it's equivalent because of the return.

PR #501, opened 2018-08-30
Add sender type to server_command event

Pieter12345:
Add sender type to the server_command event. This can be one of: console, command_block, command_minecart of null when CommandHelper does not know the command sender implementation.

NOTE: I am unsure about the "type" key. This could also be "sender" or "sender_type" to be more clear.


Pieter12345:
Closed due to the `string-ids` branch being merged with `master`.

PR #506, opened 2018-10-01
 Support $ variable in DoubleQuotedStrings

EntryPointKR:
We can use $ variable in DoubleQuotedStrings?


LadyCailin:
I have mixed feelings here. The $ var syntax is unique to CH, and I’m hesitant to bring that in so deeply to MethodScript. Having said that, cmdline scripts do use $1 for input parameters. 

Anyways, I’m hesitant, but open to opinions.

LadyCailin:
I think I’m going to reject this change, at least for now. I think the usage would be rare enough that it would just end up being more hassle than it’s worth, because then you’d need to escape $ if you needed a literal one, and furthermore, this would break existing scripts that have $ in the string. 

Thanks for the effort though!

PR #509, opened 2018-10-16
Add some events to CommandHelper

SilvLightWolf:


LadyCailin:
Thank you for your submission! I did a quick review, and in principal, this looks ok, though we need to go through it in detail first. But before we do that, it would be great if you could go through and fix the checkstyle problems first, that way we don't have to review it twice.

PR #511, opened 2018-11-17
Construct to mixed2018

LadyCailin:
This set of changes completes the goals of the original construct-to-mixed branch. Function unfortunately worked with Constructs, rather than the Mixed interface. This was a violation of the Dependency Inversion Principal, which has prevented several critical improvements to the language. The problem is that in Java, there is not multiple inheritance, so while normal classes could easily extend Construct and then be added into the ecosystem, for other types of classes, where it's impossible to extend Construct (particularly because they are already extending another class, which is true in at least the cases of enums and exceptions), these objects simply could not be added into the ecosystem.

Now, everything accepts and for the most part deals with Mixed, which is an interface. This has the unfortunate side effect of breaking binary compatibility with extensions, but this change has been in the works for at least three years. As part of this change, implementing enums was used as a test case. That has now also been done, and barring a few bugs, seems to mostly work. This allows access of enums as first class objects, rather than as arrays of strings as previously implemented through reflect_pull.

The PR opens lots of possibilities in the future, and so is a welcome change to the base, but there are a few short term goals after this PR is merged:

- Move exceptions into first class status. Currently they are implemented as arrays, but this is a shoehorn and not ideal.
- Write up documentation for Enums. There is some documentation already available, and this may be mostly correct, but it should at least be read through to ensure it is correct, and then linked to the learning trail. It's also possible to iterate enums with foreach, and so this should be added as an example in various places.
- There are still a few bugs with enums, particularly it is not possible to assign an enum class to a variable. ClassType @s = ArraySortType; should work, but @s remains null. (The same is true even if @s is auto typed). 
- Continue with object oriented notation. This provides a good framework to begin with. While it is possible to do EnumType[0] or EnumType['EnumName'], this is the reflective access mechanism, it should also be possible to access the enum by name like EnumType::EnumName.

Since this breaks extensions, the version number has been bumped to 3.3.4, and this PR should not be merged without coordination with extension authors. Also, before merging this PR, one more change should be made, which is to deprecate concatenation of bare strings.


src/main/java/com/laytonsmith/core/Script.java
```diff
@@ -884,7 +889,8 @@ public void checkAmbiguous(List<Script> scripts) throws ConfigCompileException {
 						// the same argument position.
 						if(c1.getCType() == c2.getCType()
 								&& (c1.getCType() == ConstructType.STRING || c1.getCType() == ConstructType.COMMAND)) {
-							if(c1.nval() != c2.nval() && (c1.nval() == null || !c1.nval().equals(c2.nval()))) {
+							if(Construct.nval(c1) != Construct.nval(c2) && (Construct.nval(c1) == null
+									|| !Construct.nval(c1).equals(Construct.nval(c2)))) {
```
LadyCailin: I just realized this is a bug, since nval can return java null, and we're dereferencing it here.

src/main/java/com/laytonsmith/core/constructs/NativeTypeList.java
```diff
@@ -129,7 +269,25 @@ public static String resolveType(String simpleName) {
...
+	/**
+	 * For documentation and other purposes, it is useful to have a default, invalid instance of the underlying type.
+	 * For enums and non-enums, the process of instantiating the value is different (MEnums are actually valid as is).
+	 *
+	 * This method abstracts away the difference, and you can call this instead.
+	 * @param fqcn
+	 * @return
+	 * @throws java.lang.ClassNotFoundException
+	 */
+	public static Mixed getInvalidInstanceForUse(FullyQualifiedClassName fqcn) throws ClassNotFoundException {
```
LadyCailin: TODO: It may be useful for some future types to provide their own implementation of this method. We should check for the existence of a particularly named static method within the class that provides a default instance, and we first check for that, and if it exists, call that instead of instantiateUnsafe.

src/main/java/com/laytonsmith/core/Script.java
PseudoKnight: Should these Constructs ever be CNull anyway?

src/main/java/com/laytonsmith/core/constructs/NativeTypeList.java
LadyCailin: Fixed

PR #526, opened 2019-05-31
Tipped arrow update (Fix ClassCastException in ProjectileLaunchEvent)

Pieter12345:
CraftTippedArrow in CraftBukkit changed from implementing TippedArrow to implementing Arrow. This causes an incompatibility that is fixed in this PR.

Bukkit changes for reference:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/c7c4cb533f426888e9eb4f1e07a6b9bb2236f476#src/main/java/org/bukkit/entity/TippedArrow.java

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCTippedArrow.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/entities/BukkitMCTippedArrow.java
@@ -3,7 +3,7 @@ package com.laytonsmith.abstraction.bukkit.entities;
 import com.laytonsmith.abstraction.MCLivingEntity;
 import com.laytonsmith.abstraction.MCPotionData;
 import com.laytonsmith.abstraction.bukkit.BukkitMCPotionData;
-import com.laytonsmith.abstraction.entities.MCTippedArrow;
+import com.laytonsmith.abstraction.entities.MCArrow;
 import com.laytonsmith.abstraction.enums.bukkit.BukkitMCPotionEffectType;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.TippedArrow;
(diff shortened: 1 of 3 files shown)


PseudoKnight:
No. I intentionally left MCTippedArrow in for backwards compat. This needs to work on 1.13 and 1.14.

PseudoKnight:
The tipped arrow nonsense was changed twice in spigot, but I thought I fixed it recently. My initial thought is that I think that cast exception bug needs to be fixed in BukkitMCEntityType.

PseudoKnight:
What's weird is I remember testing projectile_launch event bind... maybe it was projectile_hit, though. That'd be extra weird if it was different, though.

PseudoKnight:
Oh okay, I found where the problem was. It only happens if you set the potionmeta for a tipped arrow in 1.14. This was because I missed a cast to MCTippedArrow I should have removed on the last commit. (entity_managerment.class:2292)

PseudoKnight:
Unless you found a separate issue somewhere else.

Pieter12345:
Feel free to fix it any other way, but since 30-04-2019, CraftTippedArrow cannot be cast to TippedArrow anymore as it no longer implements it. This leaves 1.14 with a CraftTippedArrow that has a set of methods defined in Arrow and 1.13 with a CraftTippedArrow that has those methods defined in TippedArrow (extends Arrow). So for both versions, the Arrow class can be used, but 1.13 probably needs some reflection to access those methods without casting to TippedArrow.

PseudoKnight:
For now I'm just going to fix this in set_entity_spec(). The abstraction does work, and while using reflection here is probably possible, it does offer some challenges for a result I'm not even sure I prefer. I might revisit it later when I"m not miserably sick. There are more important things, though.

PR #530, opened 2019-07-02
Fix void handling for proc return types and assignment

PseudoKnight:
The tests should make it obvious what I was going for here, but I'm not sure this is entirely consistent with your plans. void is a bit weird.

Also, there's a bit of hard-to-read code that I cleaned up in the latter part of getProcedure() too. So, if you're wondering what I changed there, it does exactly the same thing. It's just easier to understand now.


PseudoKnight:
Brief explanation: void wasn't being set as the return type for procedures, so this needed to be fixed in Compiler.__autoconcat__.optimizeSpecial() -- because void is a very special boy -- as well as propagated in proc.getProcedure(). I took the opportunity to throw in a proper compile exception if someone tries to assign a variable to a void type.

PR #531, opened 2019-07-27
Fix numArgs to allow set_peffect icon parameter usage.

bardiharborow:


LadyCailin:
Good catch!

PseudoKnight:
Huh. I was sure I tested this when I added it, but I guess not. Maybe I just tested the meta. Thanks for the fix. There's actually another place this needs to be fixed too.

PR #532, opened 2019-07-30
Recognize Wandering Traders as Merchants

jb-aero:


src/main/java/com/laytonsmith/core/functions/Trades.java
```diff
@@ -198,7 +198,9 @@ public String getName() {
 		@Override
 		public String docs() {
 			return "array {} Returns an array where the keys are currently registered merchant IDs and the values are"
-					+ " the corresponding window titles of those merchants.";
+					+ " the corresponding window titles of those merchants. For entity merchants,"
+					+ " this will be the custom name if it has one, falling back to Profession for Villagers,"
+					+ " and entity type for any other trading entity.";
 		}
```
PseudoKnight: I'm confused about this. As far as I can tell, virtual merchants are never entity merchants. So what is this new documentation describing?

jb-aero:
Maybe I should note this specifically. Virtual merchants have their title
set when they are created, which gives us a guarantee of what the title is.
There's no api for getting the actual window title so for entities we have
to just do our best if it doesn't have a custom name.

PseudoKnight:
So you're describing what the titles will be when not using virtual merchants? It's just confusing that it's under the get_virtual_merchants() function docs.

jb-aero:
Ah. Yes, that is indeed mis-documentation. A virtual merchant is never an
entity.

PR #534, opened 2019-08-05
Fix unknown target in CompositeFunction

EntryPointKR:
You can easily reproduce this problem using below code
```
@a = null
broadcast(string_contains(@a, 't'))
```
Then CommadHelper will print unknown source of error.
So I tried simple way to fix that. 
But I'm not sure that is this correct way. Please review this.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/CompositeFunction.java
+++ b/src/main/java/com/laytonsmith/core/functions/CompositeFunction.java
@@ -51,7 +51,6 @@ public abstract class CompositeFunction extends AbstractFunction {
 				tree = MethodScriptCompiler.compile(MethodScriptCompiler.lex(script, null, true), environment)
 						// the root of the tree is null, so go ahead and pull it up
 						.getChildAt(0);
-				setTarget(tree, t);
 			} catch (ConfigCompileException | ConfigCompileGroupException ex) {
 				// This is really bad.
 				throw new Error(ex);
@@ -62,6 +61,7 @@ public abstract class CompositeFunction extends AbstractFunction {
 		} else {
 			tree = CACHED_SCRIPTS.get(this.getClass());
 		}
+		setTarget(tree, t);
 
 		GlobalEnv env = environment.getEnv(GlobalEnv.class);
 		IVariableList oldVariables = env.GetVarList();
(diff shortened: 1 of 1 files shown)


src/main/java/com/laytonsmith/core/functions/CompositeFunction.java
```diff
@@ -54,6 +61,7 @@ public final Mixed exec(Target t, Environment environment, Mixed... args) throws
 		} else {
 			tree = CACHED_SCRIPTS.get(this.getClass());
 		}
+		setTarget(tree, t);
```
LadyCailin: Hmm, this will cause issues if there are multithreaded uses of the function, so I think this is not a good idea to do this. Additionally, the setTarget function is going to be quite inefficient, because every call requires O(n) processing of the whole tree. I'd rather just catch an exception that the eval throws, and then set the target accordingly. Then we are limited to changing things that only affect this thread, and we avoid having to loop through all the nodes in the tree.

EntryPointKR: Nice idea

LadyCailin: Actually, this is more complicated, because the exception stack trace needs to be modified, at the correct spot, because lower stack entries may already have a target that's correct, so it's not a matter of just setting the correct target on the exception itself, it's about setting the target of the stack trace element. I would need to dig further into this to figure out exactly how to do that, but if you want to do it, you'll need to examine the exception's stack trace (which is actually stored in the environment, not the exception), and make the appropriate change there.

src/main/java/com/laytonsmith/core/functions/CompositeFunction.java
```diff
@@ -78,6 +70,10 @@ public final Mixed exec(Target t, Environment environment, Mixed... args) throws
 			}
 		} catch (FunctionReturnException ex) {
 			ret = ex.getReturn();
+		} catch (ConfigRuntimeException ex) {
+			ex.setTarget(t);
```
LadyCailin: Only set this if the current target is UNKNOWN. I think.

PR #536, opened 2019-08-08
Add all_materials() function

EntryPointKR:
I really worried about where `getMaterialValues()` should be placed.
I refered `getEnchantmentValues` so I placed it in BukkitConverter.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/Convertor.java
+++ b/src/main/java/com/laytonsmith/abstraction/Convertor.java
@@ -47,6 +47,8 @@ public interface Convertor {
 
 	void Startup(CommandHelperPlugin chp);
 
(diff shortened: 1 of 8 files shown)


EntryPointKR:
My IDE did repositioning imports automatically.  I tried to revert them but it seems not bad. Should I revert?

PseudoKnight:
The import repositioning is fine in this case.

One suggestion: add isLegacy() to MCMaterial and use that instead of startsWith("LEGACY_"). This should improve operation time.

PseudoKnight:
I'd prefer to filter out legacy materials in the GetMaterialValues(), if you want to do that instead.

EntryPointKR:
Thats nice

PR #537, opened 2019-08-09
The null should be not contain in get_all_recipes()

EntryPointKR:
If a recipe is not supported type, returns null. [Check it](https://github.com/EngineHub/CommandHelper/blob/89017ed2b790c5847c8e54438f1b666510290826/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitConvertor.java#L678)
There is unsupported types: `BlastingRecipe, CampfireRecipe, SmokingRecipe, StonecuttingRecipe`

Is there a case for null checking to how many types unsupported in CH? I don't think that


PseudoKnight:
Thanks for bringing this up. I've been working on other aspects of recipes, and I think I'll fix this in a different way.

EntryPointKR:
Cool!

PR #539, opened 2019-08-22
Avoid unnecessary hashing when getting list of types

PseudoKnight:
ArrayList is much faster here, since we generally want to loop through this list anyway. It also maintains the SortedSet order, if that's useful. It seems intentional to use SortedSet there, though I'm not sure why type unions need an order.

I mentioned this yesterday, but I'm not sure you caught it, since you were focused on caching. I just did some testing today and it looks good. I saw 10-19% reductions in total operation times. (~13% avg)

(BTW, I did 5 runs per build, and the runs are around 30 seconds. It loops through a compilation of various operations I've measured over the years, though I've removed IO to reduce variation. It doesn't count compilation time.)

### DIFF
--- a/src/main/java/com/laytonsmith/core/constructs/CClassType.java
+++ b/src/main/java/com/laytonsmith/core/constructs/CClassType.java
@@ -17,6 +17,7 @@ import com.laytonsmith.core.objects.ObjectDefinitionNotFoundException;
 import com.laytonsmith.core.objects.ObjectDefinitionTable;
 import com.laytonsmith.core.objects.UserObject;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -455,7 +456,7 @@ public final class CClassType extends Construct implements com.laytonsmith.core.
 	 * @return
 	 */
 	protected Set<CClassType> getTypes() {
-		Set<CClassType> t = new HashSet<>();
+		Set<CClassType> t = new TreeSet<>(Comparator.comparing(CClassType::getFQCN));
 		for(FullyQualifiedClassName type : types) {
 			try {
 				t.add(CClassType.get(type));


LadyCailin:
I don't think I like this. It may be that HashSet is slower, but you've changed the semantics here by changing from Set to List. A Set is not allowed to have duplicate values, while Lists are. I think the better solution here is to see about the performance of TreeSet, and see if that's more comparable to ArrayList. Insertion will always be slower in a Set though, because it has to check for duplicates somehow or another O(n) at worst, O(log n) at best, whereas a list just tacks it on to the end, which is O(1). However, the TreeSet can be provided a custom comparator, which can just use the fully qualified name of the class instead, which should at least avoid the hashing.

PseudoKnight:
I thought about redundancies, but I don't know why it would be an issue as the code stands. If a type union is created with multiples of the same type, that can be checked at compile time. Even if it isn't, it's just an inefficiency in the user's code, right?

PseudoKnight:
So I tried TreeSet without using a custom comparator. It took over 3x as long. Whoops.

After the custom comparator, if anything it's faster than the ArrayList. Nice. Good call.

PseudoKnight:
How's that?

src/main/java/com/laytonsmith/core/constructs/CClassType.java
```diff
@@ -454,8 +455,8 @@ public boolean unsafeIsExtendedBy(CClassType checkClass) {
 	 *
 	 * @return
 	 */
-	protected Set<CClassType> getTypes() {
-		Set<CClassType> t = new HashSet<>();
+	protected TreeSet<CClassType> getTypes() {
```
LadyCailin: [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle) is the preferred approach here, so still want to return Set, as well as use Set on the lhs of the assignment.

PseudoKnight: Oh right. I usually do that but I forget sometimes.

PR #541, opened 2019-09-01
Support executing closures in variables using function syntax.

PseudoKnight:
This is an early implementation that makes `@closure(@arg1, @arg2)` equivalent to `execute(@arg1, @arg2, @closure)`. Also fixes LangServ incorrectly checking procedures as functions.

While this is obviously not how you eventually plan to add this, this does add initial support for this concept and makes proc checking more clear. This minimal modification implementation should be just as easily updated with whatever you have planned.

This obviously doesn't support `@array['closure'](@arg1, @arg2)`, which will require more work.

### DIFF
--- a/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
+++ b/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
@@ -1824,8 +1824,8 @@ public final class MethodScriptCompiler {
 			//Don't care about these
 			return;
 		}
-		if(tree.getData().val().startsWith("_")) {
-			//It's a proc. We need to recurse, but not check this "function"
+		if(!((CFunction) tree.getData()).hasFunction()) {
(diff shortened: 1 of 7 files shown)


LadyCailin:
This looks good, but can you add some unit tests for it? We want to ensure that it optimizes down the the correct format.

PseudoKnight:
This doesn't optimize down to anything (at least intentionally) like execute(). We could do it that way, though. We could even "optimize" procs to call_proc().

LadyCailin:
Oh, yeah, that's a good point. But either way, a unit test or two would be good.

This is *probably* more efficient, but either way, we can keep the implementation this way.

PseudoKnight:
Okay, I'll do that. (not tonight)

I do think this is more efficient too, at least for just an ivariable. Will probably have to compile the array closure method to execute(), I imagine.

PR #545, opened 2019-12-27
Update PlayerManagement.java

jjkrause:
Pfacing() documentation was incorrect.  It indicated yaw 0 is dead west and 90 is north, but all the testing I'm doing is showing 0 as south and 90 west, etc.  Same for pfacing(F).  It says 0-West, 1-South, 2-East, 3-North.  My testing shows F0=South, F1=West, F2=North and F3=East.


PseudoKnight:
This is because Mojang changed cardinal directions eight years ago and nobody updated the documentation until now.

Issue #1180, opened 2020-04-25
DuplicateObjectDefintionException.java Is misspelled


LadyCailin:
Pls fix

PR #1181, opened 2020-04-26
fix: Fix bug #1180 misspelled class name

acasadoquijada:
The DuplicateObjectDefintionException.java class is misspelled,
it should be DuplicateObjectDefinitionException.java

This commit fixes the class name.

Resolves: #1180


LadyCailin:
Haha, this was more a note for myself to fix when I wasn't on my phone, but thanks!

PR #1183, opened 2020-05-03
Fix cmdline file access security checks

Pieter12345:
- Bypass Security.CheckSecurity(File) checks in cmdline mode.
- Move existing in-cmdline-mode checks before Security checks, since Security checks can throw an IOEx that shouldn't have to occur in cmdline mode.


LadyCailin:
Checkstyle failures 😬

Pieter12345:
> 
> 
> Checkstyle failures 😬

Those are solved in my other PR, they are currently present on master.

PR #1187, opened 2020-05-13
Add remove_recipe().

Lildirt:


Lildirt:
The function would return false if the key didn't exist. The Bukkit method does that part for us.

LadyCailin:
Oh, I see that now. The documentation is a bit confusing. Maybe clarify what the return value means.

PR #1198, opened 2020-05-29
Static analysis

Pieter12345:
Add more static analysis to MethodScript. This includes:
- Detecting undefined variables.
- Detecting undefined procedures.
- Detecting duplicate variable declarations.
- Type checking framework.
- Type checking BasicLogic functions.
- Type checking assign().
- Type checking a handful of ArrayHandling functions.

### DIFF
--- a/src/main/java/com/laytonsmith/core/Static.java
+++ b/src/main/java/com/laytonsmith/core/Static.java
@@ -46,6 +46,7 @@ import com.laytonsmith.core.constructs.Variable;
 import com.laytonsmith.core.environments.CommandHelperEnvironment;
 import com.laytonsmith.core.environments.Environment;
 import com.laytonsmith.core.environments.GlobalEnv;
+import com.laytonsmith.core.environments.RuntimeMode;
 import com.laytonsmith.core.exceptions.CRE.CREBadEntityException;
 import com.laytonsmith.core.exceptions.CRE.CRECastException;
 import com.laytonsmith.core.exceptions.CRE.CREFormatException;
@@ -79,6 +80,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLEncoder;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -1304,15 +1306,14 @@ public final class Static {
 	 * Generates a new environment, assuming that the jar has a folder next to it named CommandHelper, and that folder
 	 * is the root.
 	 * @param install
-	 * @param inCmdlineMode - {@code true} to generate a cmdline environment, {@code false} otherwise.
-	 * @param inInterpreterMode - {@code true} to generate an intepreter environment, {@code false} otherwise.
+	 * @param runtimeModes The {@link RuntimeMode}s for this environment.
 	 * @return
 	 * @throws IOException
 	 * @throws DataSourceException
 	 * @throws URISyntaxException
 	 */
 	public static Environment GenerateStandaloneEnvironment(
-			boolean install, boolean inCmdlineMode, boolean inInterpreterMode)
+			boolean install, EnumSet<RuntimeMode> runtimeModes)
 			throws IOException, DataSourceException, URISyntaxException, Profiles.InvalidProfileException {
 		File platformFolder = MethodScriptFileLocations.getDefault().getConfigDirectory();
 		if(install) {
@@ -1328,14 +1329,14 @@ public final class Static {
 				new URI(URLEncoder.encode("sqlite://" + new File(platformFolder, "persistence.db").getCanonicalPath().replace('\\', '/'), "UTF-8")), options);
 		GlobalEnv gEnv = new GlobalEnv(new MethodScriptExecutionQueue("MethodScriptExecutionQueue", "default"),
 				new Profiler(MethodScriptFileLocations.getDefault().getProfilerConfigFile()), persistenceNetwork,
-				platformFolder, profiles, new TaskManagerImpl(), inCmdlineMode, inInterpreterMode);
+				platformFolder, profiles, new TaskManagerImpl(), runtimeModes);
 		gEnv.SetLabel(GLOBAL_PERMISSION);
 		return Environment.createEnvironment(gEnv, new CompilerEnvironment());
 	}
 
 	/**
 	 * Generates a new environment, assuming that the jar has a folder next to it named CommandHelper, and that folder
-	 * is the root. This new environment is not in cmdline and interpreter mode.
+	 * is the root. This new environment is in embedded mode (and not in cmdline and interpreter mode).
 	 * @param install
 	 * @return
 	 * @throws IOException
@@ -1344,12 +1345,12 @@ public final class Static {
 	 */
 	public static Environment GenerateStandaloneEnvironment(boolean install)
 			throws IOException, DataSourceException, URISyntaxException, Profiles.InvalidProfileException {
-		return GenerateStandaloneEnvironment(install, false, false);
+		return GenerateStandaloneEnvironment(install, EnumSet.of(RuntimeMode.EMBEDDED));
 	}
 
 	/**
 	 * Generates a new environment, assuming that the jar has a folder next to it named CommandHelper, and that folder
-	 * is the root. This new environment is not in cmdline and interpreter mode.
+	 * is the root. This new environment is in embedded mode (and not in cmdline and interpreter mode).
 	 * @return
 	 * @throws IOException
 	 * @throws DataSourceException
--- a/src/main/java/com/laytonsmith/core/Static.java
+++ b/src/main/java/com/laytonsmith/core/Static.java
@@ -1303,13 +1303,17 @@ public final class Static {
 	/**
 	 * Generates a new environment, assuming that the jar has a folder next to it named CommandHelper, and that folder
 	 * is the root.
-	 *
+	 * @param install
+	 * @param inCmdlineMode - {@code true} to generate a cmdline environment, {@code false} otherwise.
+	 * @param inInterpreterMode - {@code true} to generate an intepreter environment, {@code false} otherwise.
 	 * @return
 	 * @throws IOException
 	 * @throws DataSourceException
 	 * @throws URISyntaxException
 	 */
-	public static Environment GenerateStandaloneEnvironment(boolean install) throws IOException, DataSourceException, URISyntaxException, Profiles.InvalidProfileException {
+	public static Environment GenerateStandaloneEnvironment(
+			boolean install, boolean inCmdlineMode, boolean inInterpreterMode)
+			throws IOException, DataSourceException, URISyntaxException, Profiles.InvalidProfileException {
 		File platformFolder = MethodScriptFileLocations.getDefault().getConfigDirectory();
 		if(install) {
 			Installer.Install(platformFolder);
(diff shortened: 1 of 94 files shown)


src/main/java/com/laytonsmith/core/functions/BasicLogic.java
```diff
@@ -147,6 +152,15 @@ public CBoolean exec(Target t, Environment env, Mixed... args) throws CancelComm
...
 		}
 
+		@Override
+		public CClassType getReturnType(Target t, List<CClassType> argTypes,
+				List<Target> argTargets, Environment env, Set<ConfigCompileException> exceptions) {
+			for(int i = 0; i < argTypes.size(); i++) {
+				StaticAnalysis.requireType(argTypes.get(i), Mixed.TYPE, argTargets.get(i), env, exceptions);
+			}
+			return CBoolean.TYPE;
+		}
```
LadyCailin, line 162: I kind of would rather invert this, and simply have the function return its types, and then the static analysis tool determines the correct signature and validates against that. So getReturnType would just return CBoolean.TYPE, and that's it in this case. Then a separate method would return the argument types.

This work was begun in another branch, NewCompiler, and there's several library classes for generating this information, which we can cherry pick and add to this branch. See https://github.com/EngineHub/CommandHelper/tree/NewCompiler/src/main/java/com/laytonsmith/core/arguments and see https://github.com/EngineHub/CommandHelper/blob/NewCompiler/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java#L86 for an example use. 

For some functions, this is not sufficient, and we may very well need more complex logic like you have implemented here, and that's fine for a more complex function, but using this method for everything will end up with a TON of duplicated code, varying only slightly.

Pieter12345: I currently have two typecheck functions which can be overridden. One gets the AST terms, scope graph and anything it might need access to, and the default implementation gets the types from the arguments and passes them to the getReturnType() we see here. What we could do is change the default implementation of getReturnType() to do even lower level logic, which would indeed be fine for many simple functions. That way, there are three levels at which you can implement typechecking: Very little control, reasonable conditional typing control and full AST and scope graph access control.

LadyCailin: Yes, that seems reasonable. Actually, I think in almost all cases, we would want to implement this new method anyways, for documentation and reflective purposes, so we don't generally need an AST to figure out what arguments and return types are, so that probably would be required for all functions anyhow.

LadyCailin: The more I think about it, the more I think that's definitely true. We can use the overridable methods as a stop gap, but we do need to get to the point where we are formally and statically able to declare the signature of functions. And actually, to do that, we need to probably combine the return type and input parameters into one, so we can have overloaded signatures with varying return types based on parameters, so different than how I had done it in that other branch as well. Need to return a single Signature object which contains all signatures. Given the scope of this change, I think probably this should be kept out of scope for this PR, and we can revisit that once we get this all working. Good that you have validated that in general you're able to do this with the static analysis though.

Pieter12345: It would be great if documentation is generated from the same code that is used in typechecking, since then the documentation and implementation would always match. It'd be nice if you could complete that system and write a default getReturnType() implementation in AbstractFunction that calls that. I do see that you are using for example CInt.class where I use CInt.TYPE. I think it would be good to use CClassTypes everywhere to make all type handling in compile time work on the same system.

LadyCailin: Yes, CClassType was introduced after this branch. But user classes are not going to be representable through java classes, so this change was inevitable. But the principal is mostly the same.

Pieter12345: Considering that adding some signature to functions and using those to get both the return type and generate documentation can be done using the current implementation in the PR, and that it is out of scope for this PR, I'm marking the conversation as resolved here.

Pieter12345:
Rebased on master to not mix up all commits.

src/main/java/com/laytonsmith/core/functions/ControlFlow.java
```diff
@@ -117,6 +116,26 @@ public Mixed exec(Target t, Environment env, Mixed... args)
...
+				ParseTree ast, Environment env, Set<ConfigCompileException> exceptions) {
+			if(ast.numberOfChildren() < 2) {
+				return parentScope;
+			}
+
+			// Handle condition in parent scope.
+			Scope condScope = analysis.linkScope(parentScope, ast.getChildAt(0), env, exceptions);
+
+			// Handle if and else branches in separate scopes.
+			analysis.linkScope(analysis.createNewScope(condScope), ast.getChildAt(1), env, exceptions);
```
LadyCailin, line 130: I still would really like to make this a bit more inverted. Instead of passing the analysis in and doing the work manually each time, it would be better for each function to report which scopes each branch has, and then the engine do the actual work of calling linkScope. Where absolutely necessary, this can be written like this, but I think there should be another abstraction layer on top of this that builds off of the VariableScope mechanism.

In any case, there are now two different mechanisms in the code base that do effectively the same work, so one of them should definitely be removed.

src/main/java/com/laytonsmith/core/functions/DataHandling.java
```diff
@@ -2040,6 +2353,54 @@ public Mixed execs(Target t, Environment env, Script parent, ParseTree... nodes)
...
+				Environment env, Set<ConfigCompileException> exceptions, boolean codeInheritsParentScope) {
+
+			// Handle empty closure.
+			if(ast.numberOfChildren() == 0) {
+				return parentScope;
+			}
+
+			// Handle optional return type argument.
+			int ind = 0;
+			Scope retTypeScope = (ast.getChildAt(ind).getData().isInstanceOf(CClassType.TYPE)
```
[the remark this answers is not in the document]
Pieter12345: Some code could be removed there. That term should always be a CClassType and otherwise generate an error elsewhere (could be during typechecking, but the rewrite step after parsing might be better, if we consider this a syntax error). In case it is not a CClassType, someone could have any code here. Not using this scope just means that whatever you declare in that term won't be usable after that term.

src/main/java/com/laytonsmith/core/functions/DataHandling.java
```diff
@@ -2040,6 +2353,54 @@ public Mixed execs(Target t, Environment env, Script parent, ParseTree... nodes)
...
+
+			// Insert @arguments parameter.
+			paramScope.addDeclaration(new ParamDeclaration("@arguments", CArray.TYPE, ast.getTarget()));
+
+			// Handle closure parameters from left to right.
+			Scope valScope = parentScope;
+			while(ind < ast.numberOfChildren() - 1) {
+				ParseTree param = ast.getChildAt(ind++);
+				Scope[] scopes = analysis.linkParamScope(paramScope, valScope, param, env, exceptions);
+				valScope = scopes[1];
```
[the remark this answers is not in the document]
Pieter12345: So when you do:
`closure(@a = (@b = 2), @c = @b)`
Then passing `valScope` to every next child allows usage of declarations in the previous parameter value. However, these variables are not available after the entire `closure()` call, so the last `valScope` doesn't get used. `paramScope` is the scope in which the parameter declarations have been made, but not any declarations in their default values.

src/main/java/com/laytonsmith/core/compiler/analysis/StaticAnalysis.java
```diff
@@ -0,0 +1,921 @@
...
+	 * @param valScope - The scope to which a new scope is linked in which the assigned value will be handled.
+	 * @param ast
+	 * @param env
+	 * @param exceptions
+	 * @return The resulting scopes in format {paramScope, valScope}.
+	 */
+	/*
+	 *  TODO - Mark IVAR declarations as parameters, such that they can be shadowed and such that they can override
+	 *  variable declarations. Otherwise this will lead to false duplicate declarations.
+	 */
```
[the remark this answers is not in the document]
Pieter12345: That's implemented indeed.

src/main/java/com/laytonsmith/core/functions/DataHandling.java
```diff
@@ -318,11 +353,181 @@ public Mixed exec(Target t, Environment env, Mixed... args) throws CancelCommand
 			return v;
 		}
 
+		@Override
```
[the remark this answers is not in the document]
Pieter12345: The scope graph I've created for this analysis does not follow the control flow in a way you'd need to find the latest assign(s) of a variable. Imagine:
```
int @a = 1;
if(...) {
    @a = 2;
}
use(@a);
```
Here, usage of `@a` can really only resolve to the declaration, since the second assign is out of scope according to the scope graph. Having that said, the scope graph implementation does support creating namespace-specific scope edges, so what we can do is:
```
1. Put each variable assign as a declaration into a new namespace.
2. Add namespace-specific edges to control flow functions for this namespace, such that if/else branches are accessible after the if() call etc.
3. For each variable reference, check whether the variable assign(s) it resolves to is/are static (and have the same value in case there's multiple).
```
It's funny how I thought that this wasn't possible with this analysis, but actually it is.

src/main/java/com/laytonsmith/core/functions/ControlFlow.java
Pieter12345: There is a good default implementation for all functions that do not have to do anything with scoping, so those won't be bothered with this. As for scoping functions, I think behavior of how things depend on each other is so different accross them that adding another layer on top won't do any good. Also, when adding something like the analysis as you suggested (replacement of final values), the way scopes have to be linked might diverge even more.

Pieter12345: Looking through implementations, I realized that `for()`, `forelse()`, `dowhile()` and `while()` could just call the default implementation instead of handling their arguments in order themselves. The remaining implementations are quite divergent.

src/main/java/com/laytonsmith/core/Static.java
```diff
@@ -1287,13 +1327,36 @@ public static Environment GenerateStandaloneEnvironment(boolean install) throws
 		PersistenceNetwork persistenceNetwork = new PersistenceNetworkImpl(MethodScriptFileLocations.getDefault().getPersistenceConfig(),
 				new URI(URLEncoder.encode("sqlite://" + new File(platformFolder, "persistence.db").getCanonicalPath().replace('\\', '/'), "UTF-8")), options);
 		GlobalEnv gEnv = new GlobalEnv(new MethodScriptExecutionQueue("MethodScriptExecutionQueue", "default"),
-				new Profiler(MethodScriptFileLocations.getDefault().getProfilerConfigFile()), persistenceNetwork, platformFolder,
-				profiles, new TaskManagerImpl());
+				new Profiler(MethodScriptFileLocations.getDefault().getProfilerConfigFile()), persistenceNetwork,
+				platformFolder, profiles, new TaskManagerImpl(), inCmdlineMode, inInterpreterMode);
```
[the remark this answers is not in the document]
Pieter12345: Done.

src/main/java/com/laytonsmith/core/compiler/analysis/Declaration.java
```diff
@@ -0,0 +1,49 @@
+package com.laytonsmith.core.compiler.analysis;
+
+import com.laytonsmith.core.constructs.CClassType;
+import com.laytonsmith.core.constructs.Target;
+
+/**
+ * Represents a declaration in a scope graph.
+ * @author P.J.S. Kools
+ */
+public class Declaration {
```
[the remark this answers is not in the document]
Pieter12345: No, this is the way to go unless you want to store more information about the declaration, then you'll have to create a subclass. This class is actually instanciated directly for IVariable declarations.

src/main/java/com/laytonsmith/core/functions/Function.java
```diff
@@ -74,6 +78,62 @@
...
+	 * @param t - The code target, used for setting the code target in thrown exceptions.
+	 * @param argTypes - The types of the passed arguments.
+	 * @param argTargets - The {@link Target}s belonging to the argTypes (in the same order).
+	 * @param env - The {@link Environment}, used for instanceof checks on types.
+	 * @param exceptions - A set to which all type errors will be added.
+	 * @return The return type of this function when invoked with the given argument types. If the return type
+	 * is unknown, {@link CClassType#AUTO} is returned, indicating that this value should be handled as dynamic
+	 * during static type checking.
+	 */
+	public CClassType getReturnType(Target t, List<CClassType> argTypes,
```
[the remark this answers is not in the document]
Pieter12345: Out of scope for this PR, but necessary before implementing typechecking in all functions (to prevent double work). Once there is some signature setup and all functions return some signature, I don't mind implementing some default implementation of getReturnType() to use this signature.

src/main/java/com/laytonsmith/core/functions/Exceptions.java
```diff
@@ -504,6 +537,33 @@ public Mixed execs(Target t, Environment env, Script parent, ParseTree... nodes)
...
+				for(int i = 1; i < ast.numberOfChildren() - 1; i += 2) {
+					ParseTree exParam = ast.getChildAt(i);
+					ParseTree catchCode = ast.getChildAt(i + 1);
+					Scope[] scopes = analysis.linkParamScope(parentScope, parentScope, exParam, env, exceptions);
+					Scope exParamScope = scopes[0]; // paramScope.
+					analysis.linkScope(exParamScope, catchCode, env, exceptions);
+				}
+
+				// Handle optional last catch block.
+				if((ast.numberOfChildren() & 0x01) == 0x00) { // (size % 2) == 0.
```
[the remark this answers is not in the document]
Pieter12345: Moved 'complex' math to PureUtilities.

src/main/java/com/laytonsmith/core/compiler/analysis/StaticAnalysisConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+		return config;
+	}
+
+	private final Preferences prefs;
+
+	private StaticAnalysisConfiguration() {
+		List<Preferences.Preference> defaults = new ArrayList<>(Arrays.asList(
+			new Preference("global-enable", "on", Type.BOOLEAN, "This globally enables static analysis. If this is off,"
```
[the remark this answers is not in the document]
Pieter12345: Turned off and added a TODO to update it later.

src/main/java/com/laytonsmith/core/compiler/analysis/StaticAnalysisConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+			}
+		}
+		return config;
+	}
+
+	private final Preferences prefs;
+
+	private StaticAnalysisConfiguration() {
+		List<Preferences.Preference> defaults = new ArrayList<>(Arrays.asList(
+			// TODO - Replace "off" with "on" as soon as static analysis should be enabled by default.
```
LadyCailin, line 41: Also need to add an upgrade routine to delete the config file, otherwise the previously set value will persist. But I'll remember that when this is changed, so don't need to change the TODO.

src/main/java/com/laytonsmith/core/environments/GlobalEnv.java
```diff
@@ -84,14 +86,16 @@
 	 * @param root The root working directory to use
 	 * @param profiles The Profiles object to use
 	 * @param taskManager The TaskManager object to use
+	 * @param runtimeModes The {@link RuntimeMode}s for this environment.
 	 */
 	public GlobalEnv(ExecutionQueue queue, Profiler profiler, PersistenceNetwork network,
-			File root, Profiles profiles, TaskManager taskManager) {
+			File root, Profiles profiles, TaskManager taskManager, EnumSet<RuntimeMode> runtimeModes) {
```
LadyCailin, line 92: Nitpick that you don't need to change - this is a violation of the D in SOLID, since Set is the interface, and EnumSet is the concrete implementation, but this is fine, it will always be an EnumSet anyways.

Pieter12345: I do agree, but I also think that this ensures users to construct the intended efficient set here.

PR #1202, opened 2020-05-31
Clean log tests

Anatoliy057:
Correction of some tests so that there are no errors in the logs from static analysis, for which tests are not designed, i hope.

### DIFF
--- a/src/test/java/com/laytonsmith/core/functions/DataHandlingTest.java
+++ b/src/test/java/com/laytonsmith/core/functions/DataHandlingTest.java
@@ -70,9 +70,8 @@ public class DataHandlingTest {
 				+ " proc('_proc',"
 				+ "     msg('hello world')"
 				+ " )"
-				+ "@name_proc = '_proc'"
-				+ " msg(is_proc(@name_proc))"
-				+ " call_proc(@name_proc)"
+				+ " msg(is_proc('_proc'))"
+				+ " call_proc(dyn('_proc'))"
 				+ "<<<\n";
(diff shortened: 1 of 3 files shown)


src/test/java/com/laytonsmith/core/functions/DataHandlingTest.java
```diff
@@ -70,8 +70,9 @@ public void testCallProcIsProc() throws Exception {
 				+ " proc('_proc',"
 				+ "     msg('hello world')"
 				+ " )"
-				+ " msg(is_proc('_proc'))"
-				+ " call_proc('_proc')"
+				+ "@name_proc = '_proc'"
+				+ " msg(is_proc(@name_proc))"
```
LadyCailin, line 74: There is a function called dyn() which is meant for use in these cases, as eventually, once more static analysis is complete, this will also cause the same warning. https://methodscript.com/docs/3.3.4/API/functions/dyn.html This forces the compiler to treat the value as dynamic, even if it's hardcoded, and is the preferred solution in this case, and in the eval case as well.

Anatoliy057:
Similar [tests](https://github.com/Anatoliy057/CommandHelper/blob/70e355ffad8778d4cb368533f13841a882af58a9/src/test/java/com/laytonsmith/core/functions/MetaTest.java#L75): testEval(), testEval1().
Delete?

LadyCailin:
Nah, leave those, it's good to have a variety of different ways of doing the same thing, to help ensure better coverage in general.

PR #1208, opened 2020-06-10
Fix NPEs in CClassType

Pieter12345:
The CClassType invalidType is cached in a thread-unsafe way. This causes problems when LangServ is performing an operation on the same CClassType, since one thread can start using the Mixed[1] created by the other thread before it is populated with data, getting null when reading its first element.

### DIFF
--- a/src/main/java/com/laytonsmith/core/constructs/CClassType.java
+++ b/src/main/java/com/laytonsmith/core/constructs/CClassType.java
@@ -267,32 +267,39 @@ public final class CClassType extends Construct implements com.laytonsmith.core.
 	 * at least no way this will throw a ClassCastException, so given that, we are able to supress that exception here.
 	 */
 	private void instantiateInvalidType(Environment env) {
-		if(invalidType != UNINITIALIZED) {
+		if(this.invalidType != UNINITIALIZED) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/main/java/com/laytonsmith/core/constructs/CClassType.java
```diff
@@ -266,7 +266,7 @@ private CClassType(FullyQualifiedClassName type, Target t, boolean newDefinition
 	 * do that in the constructor, we need to lazy load it. We do take pains in the constructor to ensure that there is
 	 * at least no way this will throw a ClassCastException, so given that, we are able to supress that exception here.
 	 */
-	private void instantiateInvalidType(Environment env) {
+	private synchronized void instantiateInvalidType(Environment env) {
```
LadyCailin, line 269: This is too heavy. Use a double locking paradigm:

```java
if(invalidType == UNINITIALIZED) {
    synchronized(this) {
        if(invalidType == UNININTIALIZED) {
            // Do method body
        }
    }
}
```
This makes reads not synchronized, except the very first time this is called. But after that, it'll be non-blocking.

Pieter12345: Updated, and constructed the Mixed[] in a local variable to disallow a Mixed[1] with a non-initialized first element from being returned and still causing the same problem.

PR #1209, opened 2020-06-11
Re-use environment for compilation

Pieter12345:
This is an optimization which uses the same environment to compile multiple msa scripts, rather than creating a new environment every time. Environment creation reads and parses a few files, so this is an expensive operation.


LadyCailin:
Just to recap our discussion on discord - the solution here is to break out the expensive, relatively "static" parts of the environment into their own environment, which is only regenerated at explicit points. Then creating new Compiler/Global envs is much cheaper, and we can keep better separation of concerns there.

Pieter12345:
I've made the adjustments. The build is failing due to Bukkit adding a new method to an interface which CH doesn't implement, that's unaffected by my changes.

LadyCailin:
I fixed that yesterday, merge in from master

Pieter12345:
Rebased onto master.

PR #1214, opened 2020-07-02
More feature for multithreading

Anatoliy057:
I decide removing CREIllegalArgumentException in the x_thread_join function, since the other functions ignore the absence or replaced of threads in the map.

And in order to avoid this error exception, it is necessary to check whether the thread is live, then call x_thread_join already, but between these two actions the thread may already end.


src/main/java/com/laytonsmith/core/functions/Scheduling.java
```diff
@@ -195,12 +197,13 @@ public String getName() {
 		public String docs() {
 			return "void {seconds} Sleeps the script for the specified number of seconds, up to the maximum time limit defined in the preferences file."
 					+ " Seconds may be a double value, so 0.5 would be half a second."
+					+ " If the \"interrupted status\" is true then throw InterruptedException"
```
Pieter12345, line 200: It's the `interrupt status`, not the `interrupted status`.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+		}
+
+	}
+
+	@api
+	public static class x_thread_is_alive extends AbstractFunction {
+
+		@Override
+		public Class<? extends CREThrowable>[] thrown() {
+			return new Class[0];
```
Pieter12345, line 596: You can return `null` here instead. Same for the other new functions.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{1};
+		}
+
+		@Override
+		public String docs() {
+			return "bool {mixed thread} Tests if this thread is alive. A thread is alive if it has been started and has not yet died.";
```
Pieter12345, line 636: It's true that at this point you can pass in a `mixed`, but when typechecking for this will be implemented, it will likely only be taking a `string`. Same for the thread name in other functions.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+			return true;
+		}
+
+		@Override
+		public Boolean runAsync() {
+			return null;
+		}
+
+		@Override
+		public Mixed exec(Target t, Environment environment, Mixed... args) throws ConfigRuntimeException {
```
Pieter12345, line 660: It would be nice to have `x_interrupt()` without arguments interrupt the current thread.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{1};
+		}
+
+		@Override
+		public String docs() {
+			return "void {mixed thread} Interrupts this thread. In other words \"interrupted status\" set true."
+					+ " The \"interrupted status\" can be checked using the method: " + x_is_interrupted.class.getSimpleName() + "()"
```
Pieter12345, line 687: `interrupted status` -> `interrupt status` (same as above)

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+					+ " If this thread is blocked in an invocation of the " + Scheduling.sleep.class.getSimpleName() + "(), " + x_thread_join.class.getSimpleName() + "() methods"
```
Pieter12345, line 688: Usually when referring to a function, a static final field is created for the function name, containing `new x_thread_join().getName()`. This is because the function class and name do not strictly have to be equal. Maybe @LadyCailin has a useful opinion here :)

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+					+ " then its \"interrupt status\" will be cleared and it will receive an InterruptedException.";
+		}
 	}
 
+	@api
+	@seealso({x_interrupt.class})
+	public static class x_is_interrupted extends AbstractFunction {
+
+		private static java.lang.reflect.Method interrupt;
```
Pieter12345, line 697: It is questionable whether `x_is_interrupted(not_current_thread, true)` should be allowed or not. The method you're accessing through reflection to implement this is an unsynchronized native method that might not be thread-safe. For `x_is_interrupted(current_thread, true)`, you can use static `Thread.interrupted()`, which clears the flag.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+
+		@Override
+		public Boolean runAsync() {
+			return null;
+		}
+
+		@Override
+		public Mixed exec(Target t, Environment environment, Mixed... args) throws ConfigRuntimeException {
+			Thread th;
+			boolean clearFlag = true;
```
Pieter12345, line 726: I believe that `x_is_interrupted(thread)` should not clear the interrupt flag by default. That should prevent users from accidentally clearing it, when not explicitly passing the argument.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0, 1, 2};
+		}
+
+		@Override
+		public String docs() {
+			return "bool {mixed thread=current, bool clearFlag=true} Tests whether this thread has been interrupted."
```
Pieter12345, line 769: Giving default values like this is convenient, but not consistent with other functions. Optional arguments are put between `[]`s, and the description tells what happens when they are omitted. There are examples of this in the API.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -575,15 +577,197 @@ public String getName() {
...
+					+ " The \"interrupted status\" of the thread is cleared if @clearFlag is true, else \"interrupted status\" unaffected.";
```
Pieter12345, line 770: `interrupted` -> `interrupt` (like above).
The `@clearFlag` does not need to have a `@` prefix, to be consistent with other documentation.

src/main/java/com/laytonsmith/core/functions/Threading.java
Anatoliy057: Then it is impossible to implement the case x_is_interrupted (other_thread, true).

`interrupted()` and `isInterrupted()` methods do not synchronize native `isInterrupted(clearFlag)`

src/main/java/com/laytonsmith/core/functions/Threading.java
Anatoliy057: Validate argument for string type?

Pieter12345: There's no need to do that, the type checker will eventually do that in compile time.

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345: That's why I'm proposing to not allow that case. The `isInterrupted(true)` call in `Thread` is implicitly synchronized, as it can only be called from the current thread.

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345: Oh, as a side note, `bool` here should be `boolean`, since this is the actual type in MS.

src/main/java/com/laytonsmith/core/functions/Threading.java
Anatoliy057: Then i need to do the synchronization block on the `th`

Pieter12345: I actually think that `x_is_interrupted(not_current_thread, true)` should not be allowed since when a thread is interrupted, then that thread should deal with it. For me it doesn't make sense to have a thread clear the interrupt status of another thread. Lets also throw this at @LadyCailin to see what she has to say on this.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0, 1};
+		}
+
+		@Override
+		public String docs() {
+			return "void {string [id]} Interrupts thread with the given id. In other words \"interrupt status\" set true. If id not passed, it is be current thread."
```
Pieter12345, line 688: For syntax, put the type within the optional-argument `[]`s (e.g. `void {[string id]}`.

src/main/java/com/laytonsmith/core/functions/Scheduling.java
Pieter12345: This one hasn't been updated yet.

src/main/java/com/laytonsmith/core/functions/Threading.java
Pieter12345, line 688: This documentation reads a bit weird. I'd recommend: `void {[string id]} Interrupts the thread with the given id, or the current thread if no id is given. When a thread is interrupted, its interrupt status is set to true. This status can be checked using the <insert x_is_interrupted() here> and <insert x_clear_interrupt() here> function. Note that some blocking functions will throw an InterruptedException when the thread on which they are executed is interrupted.`

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0, 1};
+		}
+
+		@Override
+		public String docs() {
+			return "boolean {string [id]} Tests whether this thread with the given id has been interrupted."
```
Pieter12345, line 745: Same with the `{[string id]}` as above.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		}
+
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0};
+		}
+
+		@Override
+		public String docs() {
+			return "boolean {} Tests whether this thread with the given id has been interrupted."
```
Pieter12345, line 793: `the given id` no longer exists for this function.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		public Integer[] numArgs() {
+			return new Integer[]{0, 1};
+		}
+
+		@Override
+		public String docs() {
+			return "boolean {[string id]} Tests whether this thread with the given id has been interrupted."
+					+ " If id not passed, it is be current thread."
+					+ " The \"interrupt status\" of the thread is unaffected by this method."
+					+ " In case if the thread does not exist or is not alive, return false";
```
LadyCailin, line 748: "Tests whether the thread with the given id (or the current thread if none is provided) is interrupted via x_interrupt. If the thread doesn’t exist, or is not alive, false is returned."

Let's not mention the thing about it affecting the interrupt state. That's a weird java quirk that's irrelevant in MethodScript.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		}
+
+		@Override
+		public Boolean runAsync() {
+			return null;
+		}
+
+		@Override
+		public Mixed exec(Target t, Environment environment, Mixed... args) throws ConfigRuntimeException {
+			return CBoolean.get(Thread.interrupted());
```
LadyCailin, line 773: Just call interrupted here. No need to return the result, or expose the java weirdness here.

Pieter12345: @LadyCailin It could actually be useful to return that boolean. Having a clear function return whether it has actually cleared something or not allows one to in this case omit the `x_is_interrupted` call.

src/main/java/com/laytonsmith/core/functions/Threading.java
Anatoliy057: So delete all mentions of interrupt status?

LadyCailin: Just use the string I provided (but pull the function name dynamically).

src/main/java/com/laytonsmith/core/functions/Threading.java
LadyCailin: Ok, fair enough, but then the docs need to change. The primary function of this function is to clear the interrupt status, in most cases, it should be good enough to simply call it and ignore the return - after this function is called, the thread's interrupt flag will be cleared, whether this call did it, or another one.

src/main/java/com/laytonsmith/core/functions/Threading.java
```diff
@@ -586,4 +587,212 @@ public Version since() {
...
+		@Override
+		public Integer[] numArgs() {
+			return new Integer[]{0};
+		}
+
+		@Override
+		public String docs() {
+			return "boolean {} Tests whether this thread has been interrupted."
+					+ " The \"interrupt status\" of the thread is cleared by this method. In other words, if this method were to be called twice in succession, the second call would return false"
+					+ " (unless the current thread were interrupted again, after the first call had cleared its \"interrupt status\" and before the second call had examined it).";
```
LadyCailin, line 795: "Clears the interrupt status of the given thread, or the thread provided. If this call actually cleared the interrupt status, true is returned, if the thread was already not interrupted, false is returned."

Actually though, this brings up another point, why is this the only method that doesn't actually accept a thread id? Unless I'm missing some reason, it should.

Pieter12345: @LadyCailin Java only has a static method that clears the interrupt flag on the current thread if only API is considered. There is a non-synchronized private native method that can be called to clear the interrupt flag for a given thread, but I wouldn't feel comfortable depending on that. When you think about it, as soon as a thread is interrupted, that thread is specifically signalled and should be handling its own interrupt flag. I can't think of a sane usage for clearing the interrupt flag on another thread, everything I come up with fits `import()/export()` better.

LadyCailin: Hmm, fair enough. Then my provided documentation string should be slightly modified, but otherwise this seems good.

Issue #1179, opened 2020-04-23
Incorrect [Using undefined variable] error on unassigned variable declarations in autoconcat


Pieter12345:
When having variable declarations in a script where that variable might never be assigned to, then an incorrect `Using undefined variable: @var` runtime error is given if that declaration is inside an autoconcat.
Example code:
```
string @var;
msg();
```

Problem:
__autoconcat__() gets the value of declaration `string @var`, causing this runtime error.

LadyCailin:
I'm not entirely sure how to fix it, without some fundamental changes to how sconcat works. But this brings up a good point that I've been meaning to address for a while, and I think now is the time to fix it. sconcat is a perfectly valid function for a user to use directly, but when we compile autoconcat into sconcat, we lose the fidelity of knowing that actually this is a user defined function or not. If it's autoconcat generated, then it's fine for us to make various assumptions about the code, if it's user defined, we shouldn't change those assumptions. So I propose we add an `__sconcat__` version, which, if autoconcat would use sconcat, it turns it into `__sconcat__` instead, which, at execution time, does the same thing as sconcat, but if we need to do further introspection, we can. In this case, we're trying to concat an undefined variable and void, and if that were user defined, of course that's an error, and it's a perfectly good thing to have an error there. But in this case, since it's compiler generated, we should suppress the error, and actually just ignore the output of the variables in general, unless we're running in alias mode, and this is the top sconcat. This also solves the other issue where we have to exit() in cmdline, or else we get the variables output to screen.

We could also start identifying statements then, for instance, if we have `__sconcat__(assign(...), msg())` then we can safely say that this isn't meant to be concatenated, because we can mark assign and msg as things that are individually statements, and then we can say that the whole chain isn't an actual sconcat. Though.. maybe that would be more difficult, because what you actually need is to track the return value, which in this case would be void (or `nothing`) but you need to error out if that's being used which is a hard concept to define.

PseudoKnight:
This is essentially "fixed" when using statements after this commit: https://github.com/EngineHub/CommandHelper/commit/bd725d0a270abc11c87aee7625cbc70e4cd4c247 . As long as a semi-colon is used, there is no error. Should this be closed?

LadyCailin:
What does doing it without semicolons yield? If that works, then yes, but the bug may have shifted some.

PseudoKnight:
[...]

Issue #814, opened 2013-10-20
Unexpected extra time being taken when import used with large array in a procedure.


LadyCailinBot:
CMDHELPER-2811 - Reported by Tom.Gebbett

Here's an example that causes the problem:

```
proc('_test',
    proc('_test2',
        @before = nano_time()
        @get = import('hdd.p.player')
        @get['uzalu']['test'] = time()
        msg((nano_time() - @before) / 1000000)
    )
    @before = nano_time()
    _test2()
    msg((nano_time() - @before) / 1000000)
)
```

this outputs two numbers, being the time each part takes. The first is always around 0.02, since that's about how long the array setting takes. However, the second number is always at least 80, though there is no code between these two outputs. In my experience, the two numbers should have a difference of 0.1 at most.

It seems that modifying a big array that has been imported in a procedure causes the procedure to take a long time to do something when it ends. I guess it's unloading the array in some way, but it's never done this in previous CH versions, afaik.

If I take the code out of _test2(), then the same thing happens at the end of _test(). I've put it in a procedure to demonstrate the problem clearly.

I should mention that the value in import('hdd.p.player') is a large array with all the player data in it. It is 194450 characters long.

Just to give some context, I use import and export to cache SQL persistence retrievals, since otherwise retrieving one key takes 50+ milliseconds, rather than less than one millisecond. (i've tried playing with the persistence config and can't seem to make it cache the database.) Because this is used so often, the extra time being taken is a massive hit to my server's performance.

I'm not posting to get suggestions on how to handle persistence, i'm just reporting what seems like a bug.

PseudoKnight:
This issue was fixed with semi-colon statements as mentioned above with this commit: https://github.com/EngineHub/CommandHelper/commit/bd725d0a270abc11c87aee7625cbc70e4cd4c247

It was also previously fixed in strict mode here:
https://github.com/EngineHub/CommandHelper/commit/69f6baa63dd996484b8d46aa014ed36ce9cb0ded

Performance will still be worse with functions returning very large arrays without semi-colons in non-strict mode, but official recommended solutions exist now without workarounds.

PR #1225, opened 2020-07-22
Don't parse __autoconcat__ to sconcat in strict mode

Pieter12345:
Parse `__autoconcat__()` to the new `__statements__()` instead of `sconcat()` in strict mode. `__statements__()` takes arguments of any type and returns `void` for typechecking, so compile errors will be generated in cases where `__autoconcat__()` used to insert `sconcat()`s, or in other words, where the user has either forgotten to put some `.` concat, or where the user has made a mistake.
This change does not affect non-strict mode, as automatically inserting concats is a feature there.
Alias syntax should also remain possible in strict mode, but only when the whole alias is nicely concatenated together by the user. Inserting multiple arguments/statements will cause the code block to be interpreted as a statements block and not as an alias redirect.

### DIFF
--- a/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
+++ b/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
@@ -2119,9 +2119,13 @@ public final class MethodScriptCompiler {
 			}
 		}
 		if(root.getData() instanceof CFunction && root.getData().val().equals(__autoconcat__.NAME)) {
+
+			// In non-strict mode, let __autoconcat__ glue arguments together with sconcat.
+			boolean returnSConcat = !root.getFileOptions().isStrict();
+
 			try {
 				ParseTree ret = ((Compiler.__autoconcat__) ((CFunction) root.getData()).getFunction())
-						.rewrite(root.getChildren(), true, envs);
+						.rewrite(root.getChildren(), returnSConcat, envs);
 				root.setData(ret.getData());
 				root.setChildren(ret.getChildren());
 			} catch (ConfigCompileException ex) {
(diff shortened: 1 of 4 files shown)


Pieter12345:
This change fixes the following issues in strict mode, but does not affect them in non-strict mode:
- #1179
- #814
- (unreported) When running a script from cmdline mode, the `sconcat` of the script is printed to the console.
Note that the last issue does not collide with the feature of being able to return value from scripts. One only has to explicitly concat their script together to make it return its value to the console.

PseudoKnight:
Note that #814 already has a workaround implemented for procedures. The main sconcat is executed but not returned in procedures.

LadyCailin:
The code here looks good, but let's add something about this in the documentation, since this will break traditional use of aliases if strict mode is enabled. Probably need to make some changes in the introductory documentation to ensure that the best practice (using run()) is mentioned (though we can still show examples of the standard alias syntax).

Pieter12345:
I believe that leaves us with the following things to do/consider/realize:
- This PR only makes changes to strict mode code.
- This PR breaks `/cmd = /cmd1 arg`, and maintains support for the preferred version: `*:/cmd = run('/cmd1') \ run('/cmd2')
`, and for the explicit concat and full string versions: `/cmd = /cmd1.' arg'` and `/cmd = '/cmd1 arg'`.
- This PR maintains support for `/cmd = /cmd1 \ run('/cmd2')`, where possible arguments have to be quoted as in the first point.
- It should be considered whether we want to allow unquoted `/cmd = /cmd2` (VS `/cmd = '/cmd2'`) commands, as this might feel weird because users would only have to quote command arguments, and handle the unquoted command itself as a string.
- The [Beginner's Guide](https://methodscript.com/docs/3.3.4/Beginner's_Guide.html) should be updated to at least quote+concat the examples, and perhaps even only include the `run()` version. This is the place where users can get pushed in a certain desired direction, which might be `cmd('/cmd arg')`, but which might also be `'/cmd arg'`. I'd like to leave that choice with @LadyCailin. The documentation update can be pushed separately from this PR if you want to do it, as this PR only removes existing syntax, and doesn't add new syntax.

PR #1227, opened 2020-07-25
Unleash events

malonnnn:
I bet you this is all wrong


PseudoKnight:
The docs() is incorrect, which is not surprising since there's so little guidance for that. First braced section is prefilters. Second braced section is data. Third braced section is mutable data. The StringUtils.Join has an ending brace for the glue argument, which is wrong.

PseudoKnight:
Might also want to add "type" to the event data in evaluate().

src/main/java/com/laytonsmith/abstraction/events/MCEntityUnleashEvent.java
```diff
@@ -0,0 +1,12 @@
+package com.laytonsmith.abstraction.events;
+import com.laytonsmith.abstraction.MCEntity;
+import com.laytonsmith.abstraction.enums.MCUnleashReason;
+import com.laytonsmith.core.events.BindableEvent;
+
+public interface MCEntityUnleashEvent extends BindableEvent {
+
+		MCEntity getEntity();
```
PseudoKnight, line 8: Not sure why these have two leading tabs or why checkstyle accepts this.

src/main/java/com/laytonsmith/core/events/drivers/EntityEvents.java
```diff
@@ -2153,4 +2154,67 @@ public boolean modifyEvent(String key, Mixed value, BindableEvent event) {
...
+	public static class entity_unleash extends AbstractEvent {
+
+		@Override
+		public String getName() {
+			return "entity_unleash";
+		}
+
+		@Override
+		public String docs() {
+			return "{type: <macro> The type of entity | id: <macro> The UUID of entity}"
```
PseudoKnight, line 2168: The prefilters should be 'type' and 'reason'. The data should be 'id', 'type', and 'reason'.

malonnnn: oh right, theres never been a uuid prefilter

malonnnn: i think it clicked how this works, hopefully i just fixed it

PR #1246, opened 2020-09-24
Support RGB colours

Murreey:
Adding support for any RGB colours in `color()`, using the 1.16 `§x§r§r§g§g§b§b` format. Detects any arg that starts with a # and splits it up accordingly. Won't do anything if it's invalid hex, or not long enough.

Also added support for it to `colorize()`, given an & before the #.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
+++ b/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
@@ -125,9 +125,15 @@ public enum MCChatColor {
 	}
 
 	public static String fromRGBValue(String hexString) {
-		if(hexString.startsWith("#")) hexString = hexString.split("#")[1];
-		if(hexString.length() > 6) hexString = hexString.substring(0, 6);
-		if(!hexString.matches("(?i)^[0-9A-F]{6}$")) hexString = "ffffff";
+		if(hexString.startsWith("#")) {
+			hexString = hexString.split("#")[1];
+		}
+		if(hexString.length() > 6) {
+			hexString = hexString.substring(0, 6);
+		}
+		if(!hexString.matches("(?i)^[0-9A-F]{6}$")) {
+			hexString = "ffffff";
+		}
 		return "\u00A7x\u00A7" + String.join("\u00A7", hexString.split(""));
 	}
 
--- a/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
+++ b/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
@@ -124,6 +124,13 @@ public enum MCChatColor {
 		return CHAR_COLORS.get(code);
 	}
 
+	public static String fromRGBValue(String hexString) {
+		if(hexString.startsWith("#")) hexString = hexString.split("#")[1];
+		if(hexString.length() > 6) hexString = hexString.substring(0, 6);
+		if(!hexString.matches("(?i)^[0-9A-F]{6}$")) hexString = "ffffff";
+		return "\u00A7x\u00A7" + String.join("\u00A7", hexString.split(""));
+	}
+
 	/**
 	 * Strips the given message of all color codes
 	 *
--- a/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
+++ b/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
@@ -126,7 +126,7 @@ public enum MCChatColor {
 
 	public static String fromRGBValue(String hexString) {
 		if(hexString.startsWith("#")) {
-			hexString = hexString.split("#")[1];
+			hexString = hexString.substring(1, hexString.length());
 		}
 		if(hexString.length() > 6) {
 			hexString = hexString.substring(0, 6);
--- a/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
+++ b/src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
@@ -125,16 +125,14 @@ public enum MCChatColor {
 	}
 
 	public static String fromRGBValue(String hexString) {
-		if(hexString.startsWith("#")) {
-			hexString = hexString.substring(1, hexString.length());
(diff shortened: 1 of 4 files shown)


src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
```diff
@@ -124,6 +124,13 @@ public static MCChatColor getByChar(char code) {
 		return CHAR_COLORS.get(code);
 	}
 
+	public static String fromRGBValue(String hexString) {
+		if(hexString.startsWith("#")) hexString = hexString.split("#")[1];
```
Pieter12345, line 128: CH code style requires `if` code blocks to be within `{}`.

src/main/java/com/laytonsmith/core/functions/Echoes.java
```diff
@@ -836,6 +842,19 @@ public Mixed exec(Target t, Environment environment, Mixed... args) throws Confi
 							b.append(color.exec(t, environment, new CString(c, t)));
 							i += sl;
 						} else {
+							if (c.equals('#')) {
+								try {
+									String subsequence2 = stext.substring(i + sl, i + sl + 7);
+									if (subsequence2.matches("(?i)^#[0-9A-F]{6}$")) {
```
Pieter12345, line 848: Build error:
`src\main\java\com\laytonsmith\core\functions\Echoes.java:848: Whitespace between if/for/while and ( is not allowed. [RegexpSingleline]`.

src/main/java/com/laytonsmith/core/functions/Echoes.java
```diff
@@ -836,6 +842,19 @@ public Mixed exec(Target t, Environment environment, Mixed... args) throws Confi
 							b.append(color.exec(t, environment, new CString(c, t)));
 							i += sl;
 						} else {
+							if (c.equals('#')) {
```
Pieter12345, line 845: Build error:
`src\main\java\com\laytonsmith\core\functions\Echoes.java:845: Whitespace between if/for/while and ( is not allowed. [RegexpSingleline]`.

Pieter12345 (changes requested):
I'd like hex color support in CH. @PseudoKnight did mention hex color support before, which I believe got stuck on some design choice. He's likely got an opinion on this topic.

src/main/java/com/laytonsmith/core/functions/Echoes.java
```diff
@@ -401,6 +401,11 @@ public Mixed exec(Target t, Environment env, Mixed... args) throws ConfigRuntime
 				//If the value is empty string, set the color to white.
 				color = MCChatColor.WHITE.toString();
 			}
+
+			if(a.matches("(?i)^#[0-9A-F]{6}$")) {
```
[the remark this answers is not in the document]
Murreey: It's actually slightly different, the one in `fromRGBValue` is checking without the `#` as it's optional. Could maybe become a `.startsWith("#")` though here instead of checking the regex.

The one here and the one in `colorize()` are the same though, but I couldn't see an easy way to reduce that duplication.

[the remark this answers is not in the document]
Pieter12345: What I mean is something like:
```java
String rgbColor = MCChatColor.fromRGBValue(a);
if(rgbColor != null) {
    color = rgbColor;
}
```
As long as `fromRGBValue` returns `null` for invalid RGB/hex colors, that should work.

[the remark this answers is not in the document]
Murreey: Oh, fair enough. I've pushed an update that reduces all three into the one, and makes the `#` non optional. Makes it simpler, thanks.

src/main/java/com/laytonsmith/abstraction/enums/MCChatColor.java
```diff
@@ -124,6 +124,17 @@ public static MCChatColor getByChar(char code) {
 		return CHAR_COLORS.get(code);
 	}
 
+	public static String fromRGBValue(String hexString) {
+		if(hexString.length() > 7) {
+			hexString = hexString.substring(0, 7);
```
Pieter12345, line 129: In this case, `null` should be returned as the string isn't a valid RGB/hex color identifier. Otherwise, `color('#FFFFFFblabla')` would be accepted.

Murreey: I did it this way because `color()` does it for everything else, where it ignores extra stuff after it recognises a colour. ie `color('abcde')` will still become the 'a' code.

PseudoKnight:
My concern was that while single character color codes exist in vanilla Minecraft, hex colors do not. This uses a bukkit specific implementation. It's also not terminal compatible. 

So, Static.MCToANSIColors() needs to be updated too to strip these.

I do like the user facing solution here: the `color('#aabbcc')` format and `&#aabbcc` for colorize(). This is exactly how I wanted to add it, even if our implementations might have differed.

LadyCailin:
Would be nice to do color math to convert the hex colors to the nearest terminal color, but.. I don't know how to do that.

LadyCailin:
Could we at least define a few well know colors, and keep those? Like #ff0000 is red, for instance.

Pieter12345:
> 
> 
> Would be nice to do color math to convert the hex colors to the nearest terminal color, but.. I don't know how to do that.
From a computer's perspective, taking the smallest euclidean distance between 3D RGB colors makes sense. To get the closest color to a given color according to how humans perceive it, you'd have to convert to LAB or LUV color space and then take the smallest distance (since RGB isn't linear to the human eye, wheres LAB and LUV are). LAB is slightly better than LUV, but LUV is sometimes used because some computations/conversions with it are cheaper/easier.

LeigerGaming:
I'd love to see this in the plugin. No clue on how to do any of the color math, though - so I'm commenting in the hopes that it will restart the discussion.

LadyCailin:
Going to go ahead and merge this in, despite the fact that the terminal colors are stripped. I'll add this to the issue tracker though, no reason to further delay what's working.

PR #1247, opened 2020-09-25
Use function signatures for typechecking

Pieter12345:
This PR adds a basic function signature framework and allows for defining signatures in functions for typechecking. In the current state, any MS function signature should be creatable, but signatures with generic types are not yet handled in typechecking.

Note: As long as the [WIP] prefix is in the PR title, it should not be merged.

### DIFF
--- a/src/main/java/com/laytonsmith/core/compiler/signature/FunctionSignature.java
+++ b/src/main/java/com/laytonsmith/core/compiler/signature/FunctionSignature.java
@@ -1,0 +1,69 @@
+package com.laytonsmith.core.compiler.signature;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Represents a single function, procedure or closure signature.
+ * @author P.J.S. Kools
+ */
+public class FunctionSignature {
+
+	private final ReturnType returnType;
+	private final List<Param> params;
+	private final List<Throws> throwsList;
+
+	/**
+	 * Creates a new {@link FunctionSignature} with the given properties.
+	 * @param returnType - The function return type.
+	 * @param params - The function parameters.
+	 * @param throwsList - The list of possibly thrown exceptions by the function.
+	 */
+	public FunctionSignature(ReturnType returnType, List<Param> params, List<Throws> throwsList) {
+		this.returnType = returnType;
+		this.params = params;
+		this.throwsList = throwsList;
+	}
+
+	/**
+	 * Creates a new {@link FunctionSignature} with the given return type and no parameters
+	 * and possibly thrown exceptions.
+	 * @param returnType - The function return type.
+	 */
+	public FunctionSignature(ReturnType returnType) {
+		this(returnType, new ArrayList<>(), new ArrayList<>());
+	}
+
+	protected void addParam(Param param) {
+		this.params.add(param);
+	}
+
+	protected void addThrows(Throws throwsObj) {
+		this.throwsList.add(throwsObj);
+	}
+
+	/**
+	 * Gets the function's return type.
+	 * @return The return type.
+	 */
+	public ReturnType getReturnType() {
+		return this.returnType;
+	}
+
+	/**
+	 * Gets the function's parameters.
+	 * @return The parameters.
+	 */
+	public List<Param> getParams() {
+		return Collections.unmodifiableList(this.params);
+	}
+
+	/**
+	 * Gets the function's possibly thrown exceptions.
+	 * @return The possibly thrown exceptions.
+	 */
+	public List<Throws> getThrows() {
+		return Collections.unmodifiableList(this.throwsList);
+	}
+}
--- a/src/main/java/com/laytonsmith/core/compiler/signature/FunctionSignature.java
+++ b/src/main/java/com/laytonsmith/core/compiler/signature/FunctionSignature.java
@@ -9,7 +9,6 @@ import com.laytonsmith.PureUtilities.Common.StringUtils;
 import com.laytonsmith.core.constructs.CClassType;
 import com.laytonsmith.core.constructs.InstanceofUtil;
 import com.laytonsmith.core.environments.Environment;
-import com.laytonsmith.core.natives.interfaces.Mixed;
 
 /**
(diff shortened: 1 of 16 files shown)


Pieter12345:
Combined introduce and move commit + rebased onto master.

src/main/java/com/laytonsmith/core/compiler/signature/FunctionSignature.java
```diff
@@ -0,0 +1,175 @@
...
+	public List<Throws> getThrows() {
+		return Collections.unmodifiableList(this.throwsList);
+	}
+
+	/**
+	 * Matches the given argument types against this signature.
+	 * @param argTypes - The argument types.
+	 * @param env - The environment.
+	 * @param allowUnmatchedArgs - Whether or not to allow more arguments than the signature prescribes.
+	 * This should be {@code false} for normal functions, and {@code true} for procedures and closures.
```
LadyCailin, line 82: Let's remove this parameter, and treat it as if it's always false in the function. The way it's going to work in MethodScript is that generics can have variadic types. Read https://github.com/EngineHub/CommandHelper/blob/genericsTake2/src/main/resources/docs/Generics#L506 for more information, but in general, expect that closure's type will be defined as `class closure<Return, Params...>` and thus a `new closure<string, int, int>` will only be able to accept 2 arguments. The default for params is `auto...` and so thus is without further specification, allowed to send any number of parameters of any type, but that is still a strictly defined value.

Pieter12345: Sounds good, I've added it because it is currently supported, and while I believe that `@arguments` is a weird hack for which varargs or an array argument should be used instead, it is a breaking change if we remove it. We could think about adding a setting in file options whether or not to allow passing too many or too few arguments to procs and closures.

Pieter12345:
> A few unit tests would be great.

Added unit tests. This PR is ready to be merged in my opinion.

src/main/java/com/laytonsmith/core/compiler/signature/Param.java
```diff
@@ -0,0 +1,76 @@
...
+	 * Parameters cannot be variable and optional at the same time, as varparams already imply optionality.
+	 * @param type - The (parent) type of the parameter.
+	 * @param name - The name of the parameter.
+	 * @param isVarParam - {@code true} if the parameter is a varparam, meaning that it matches zero or more arguments
+	 * of the given type, or one argument of type {@code array<type>}. {@code false} otherwise.
+	 * Note that a varparam is only usable as type {@code array<type>}.
+	 * @param isOptional - {@code true} if the parameter is optional, {@code false} otherwise.
+	 */
+	public Param(CClassType type, String name, boolean isVarParam, boolean isOptional) {
+		assert !isVarParam || !isOptional : "A parameter cannot be variable and optional at the same time.";
```
LadyCailin, line 27: "cannot be variadic", not "variable"

Pieter12345: Fixed

src/main/java/com/laytonsmith/core/functions/AbstractFunction.java
```diff
@@ -70,13 +71,26 @@ public Mixed execs(Target t, Environment env, Script parent, ParseTree... nodes)
...
 	/**
 	 * {@inheritDoc} By default, {@link CClassType#AUTO} is returned.
 	 */
 	@Override
 	public CClassType getReturnType(Target t, List<CClassType> argTypes,
 			List<Target> argTargets, Environment env, Set<ConfigCompileException> exceptions) {
-		return CClassType.AUTO; // No information is available about the return type.
+
+		// Match arguments to function signatures if available.
+		FunctionSignatures signatures = this.getSignatures();
```
LadyCailin, line 87: Probably worth it to cache this.

src/main/java/com/laytonsmith/core/compiler/signature/SignatureBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Adds a normal function parameter. Parameters should be added from left to right.
+	 * @param paramType - The {@link CClassType} of the parameter.
+	 * @param paramName - The name of the parameter.
+	 * @param isOptional - Whether the parameter is optional or not.
+	 * @return This {@link SignatureBuilder}, for chaining builder methods.
+	 */
+	public SignatureBuilder param(CClassType paramType, String paramName, boolean isOptional) {
```
LadyCailin, line 42: Hmm. We will probably want to expand this later to add per-parameter documentation. Especially as the IDE matures, this will become even more useful. I think it's probably worth going ahead and defining this now, since it will be a HUGE hassle to have to go back *again* and backfill this.

Pieter12345: That's fair. The return type should probably also have documentation then, but `void` would be pointless to document, so there should also be a way around that. I'm wondering whether we should make per-param and return type docs mandatory, given that they can also be explained in the general function docs, which in Java often results in duplicate information. I think that it would be wise to at least add an option to add parameter and return type documentation, but not require it (yet).

LadyCailin: It's fine to accept java null, to mean "no documentation here" but it needs to be explicit.

src/main/java/com/laytonsmith/core/functions/AbstractFunction.java
Pieter12345: I've created a cache getter that will come in convenient when getting signatures for documentation and reflective access as well.

LadyCailin (approved):

src/main/java/com/laytonsmith/core/compiler/signature/SignatureBuilder.java
Pieter12345: Added per-param docs.

LadyCailin (approved):

LadyCailin (approved):

PR #1249, opened 2020-10-17
Added pattack_cooldown().

Lildirt:
Passthrough function for `HumanEntity#getAttackCooldown()`. Tested on 1.16+ only.

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCHumanEntity.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCHumanEntity.java
@@ -42,4 +42,6 @@ public interface MCHumanEntity extends MCInventoryHolder, MCLivingEntity, MCAnim
 
 	void setCooldown(MCMaterial material, int ticks);
 	//boolean setWindowProperty(MCInventoryView.Property prop, int value);
+
+	float getAttackCooldown();
 }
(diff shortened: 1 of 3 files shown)


PseudoKnight:
First, a minor note: the contributing guidelines state that "braceable statements MUST use braces". 

Next, Static.GetPlayer() can throw CRELengthException too. This is easy to miss. 

Finally, as you might have suspected, getAttackCooldown() was added in 1.15.2, so a fallback that doesn't produce a NoSuchMethodError on lower versions would be better. There's no default value for this method, and vanilla cooldowns have existed since 1.9, so I'm not sure what a good fallback would be here -- perhaps 1.0. Or maybe a CRE since this function either works completely or not. (Maybe in the future we can standardize unsupported functions with a new exception type where appropriate, similar to UnsupportedOperationException?)

Lildirt:
Ah, wasn't aware ternary counted towards that. I see where the CRELengthException is thrown and updated the PR to amend both issues.

As for the fallback, a UOE-equivalent exception feels more suitable but I don't see this happening anywhere else for similar cases in the `BukkitMC*` classes. However, on a pre-1.15 server, we would be returning outright incorrect information for the versions that had attack cooldown. Silently returning 1.0 seems like a bad idea that might lead to headaches for anyone trying to use this function especially when they don't know about this behavior. What would be appropriate to throw here?

PR #1259, opened 2021-02-17
Fix Sublime3 Syntax-Highlighting

Ecconia:
'= was recognized as keyword, instead of a quote start.
Now quotes (and comments) will be checked first, before detecting keywords.


Ecconia:
Uff, failed the name of the commit *sigh* let me fix that.

LadyCailin:
Thanks!

PR #1261, opened 2021-02-20
Add block form event

Anatoliy057:


src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
```diff
@@ -1519,4 +1519,78 @@ public boolean modifyEvent(String key, Mixed value, BindableEvent event) {
...
+					+ "{location: The location the block formed "
+					+ " | block: Type of old block state"
+					+ " | newblock: Type of new block state}"
+					+ "{} "
+					+ "{}";
+		}
+
+		@Override
+		public Driver driver() {
+			return Driver.EXTENSION;
```
PseudoKnight, line 1546: Wrong Driver.

PseudoKnight:
Gave it some thought and I don't see a whole lot of value in getting the current block for the prefilter, or even the data for that matter. If you know the new block, you can infer the old block. And worst case scenario you can still get the old block with a function. Am I wrong?

Anatoliy057:
For the prefilter of the old block, I agree, it is redundant since it can always be accurately determined by the result, but for the data the user will have to write the processing of all cases to determine the old block, and perhaps in the future there will be more of these cases.

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
Anatoliy057: Fixed

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
```diff
@@ -1519,4 +1519,75 @@ public boolean modifyEvent(String key, Mixed value, BindableEvent event) {
...
+		@Override
+		public String getName() {
+			return "block_form";
+		}
+
+		@Override
+		public String docs() {
+			return "{newblock: <string match> Type of new block state}"
+					+ "Called when a block is formed or spreads based on world conditions."
+					+ " Cancelling this event block will not be formed."
```
Pieter12345, line 1535: `Cancelling this event will cause the block to not be formed.`

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
```diff
@@ -1519,4 +1519,75 @@ public boolean modifyEvent(String key, Mixed value, BindableEvent event) {
...
+					+ "{location: The location the block formed "
```
Pieter12345, line 1536: `The location of the formed block`

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
```diff
@@ -1519,4 +1519,75 @@ public boolean modifyEvent(String key, Mixed value, BindableEvent event) {
...
+
+				return true;
+			}
+
+			return false;
+		}
+
+		@Override
+		public Map<String, Mixed> evaluate(BindableEvent e) throws EventException {
+			if(!(e instanceof MCBlockFadeEvent)) {
```
Pieter12345, line 1575: `MCBlockFadeEvent` -> `MCBlockFormEvent`
On this line, in the exception message and in the code below.

Pieter12345 (changes requested):
Looking good overall. Using the wrong event for evaluation, which leads me to believe that you havn't tested it. Testing your code yourself is a great way to get bugs out early, and you should nearly always do that.

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
Anatoliy057: Fixed

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
Anatoliy057: Fixed

src/main/java/com/laytonsmith/core/events/drivers/BlockEvents.java
Anatoliy057: Fixed

Pieter12345 (approved):

PseudoKnight:
This looks good to me.

PR #1265, opened 2021-03-10
Update array_subtract documentation for clarity

LeigerGaming:
I have updated the `array_subtract` documentation to make it easier to read.

- Separated the short summary from the detailed information
- Made use of list items so that it's not a huge wall of text


LeigerGaming:
The Snyk security issue seems to be fixed in #1215

LadyCailin:
Yes, unfortunately, the fix for that is more complicated. One of the new downstream dependencies distributes binaries in github, so pulling that code in means that building the project locally requires configuring and logging in to github from the build machine, which is a hassle and requires pipeline changes, not to mention requires a github account. So I'm just leaving that be for now, particularly when the issue it fixes doesn't even really matter for the usage we have. I should check up on it again though, and see if the problem I cited is still applicable, it has been a while.

PR #1281, opened 2021-05-26
Update to java 16

LadyCailin:
This commit only necessarily fixes the compile errors, runtime testing
needs to happen, particularly those classes that use reflection.


LadyCailin:
The failures in this build are expected, until we update the build systems to java 11. Unfortunately, that means master won't currently build either, so the last thing to do before merging is to update the build systems to use 11, then merge.

LadyCailin:
Need to remember to update the windows installer once this branch is released. The exe is up to date now, and selects java 16. I also fixed it so that it unconditionally requests to install java (with an option to skip), since the java detection doesn't seem to work very well.

PR #1283, opened 2021-06-05
Throw compile errors in strict mode when encountering illegal statements

PseudoKnight:
Due to removing auto concat from strict mode, users will get runtime exceptions in some cases where there were no errors before. So, this creates an exception at compile time instead. 

I'm aware the plan is to create return types for functions, so this is ultimately a temporary solution. It's helped me find a few in my own code. Just make sure I'm not doing something stupid with this solution.

I also updated the Strict Mode page to be up-to-date before we move to 3.3.5.

### DIFF
--- a/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
+++ b/src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
@@ -2252,6 +2252,37 @@ public final class MethodScriptCompiler {
 						}
 					}
 				}
+				// In strict mode throw compile errors when encountering child statements in function arguments where
+				// statements are not acceptable because void would be an invalid argument type. This would otherwise be
+				// a runtime error in strict mode where auto-concat is not allowed and statements are used instead.
+				// This can be removed once a more comprehensive void return type check is implemented.
+				if(tree.getFileOptions().isStrict()) {
+					if(function instanceof BranchStatement) {
+						// Only check the arguments where void is unexpected in BranchStatement functions.
+						List<ParseTree> children = tree.getChildren();
+						List<Boolean> branches = ((BranchStatement) function).isBranch(children);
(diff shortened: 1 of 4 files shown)


Pieter12345:
Can you give an example or some examples where a runtime exception is thrown? The `__statements__` function does less than what `sconcat` used to do in strict mode (e.g. the returned result is no longer used, whereas it was in sconcat). I'm unsure as to how this change would have caused new runtime exceptions.

PseudoKnight:
Unlike sconcat, __statements__ returns void. Sometimes this will result in a bug that doesn't produce an exception (eg. where `msg(colorize('abc' 'def'));` results in an empty message), but some functions may throw an exception if void is given as an argument (eg. where `@msg = colorize('abc' 'def');` results in a CastException because a variable can't be assigned to void). This catches both types of bugs.

A recent example of this happening to me just yesterday was I had a missing comma in an array. This resulted in two of the elements in the array instead being a single void. This caused a problem later when my code tried to use this array. This fix catches this mistake while I'm still in the editor.

Pieter12345:
I see, that's a fair point, even though in the past those runtime exceptions didn't occur and 'invalid' code used to just execute with undesired behavior. #1247 will partially solve that issue by performing typechecking at compile time for users who have static analysis enabled. "partially" here comes from that the error message would become something like "Type X is expected, found void instead" or so, rather than a more user friendly message. I'm in favor of adding some special detection case for `__statements__`, but I would really like to see it being implemented within the `__statements__` or `__autoconcat__` class rather than it being a special case somewhere in the compiler. I'm not sure whether it'd be possible in `__statements__` directly, as there might not be a method that provides access to the parent AST node / function. On a side note: In the ideal situation, these cases would be a syntax exception in the parser (found statement(s) where an expression was expected), but that's a tough one to implement with the current parser and autoconcat setup.

PR #1285, opened 2021-07-20
Fix GetOSBitDepth on Windows

diaoright:


LadyCailin:
Oh, haha, wow, good catch.

Issue #1282, opened 2021-05-31
Color math to get terminal colors from color(RGB)


LadyCailin:
Currently, when doing color() with an RGB value, this is simply stripped from the output. However, we should be able to do some color math to find the closest applicable terminal color, and use that instead, or at minimum translate well defined colors such as #ff0000 to "red".

See #1246 for further information, and the original PR that added this feature.

PseudoKnight:
Apparently Spigot already translates their hex code format to terminal colors when using sendMessage for ConsoleCommandSender. So we could change Static.SendMessage to use that, as long as we make sure to support `use-colors` in Preferences, as you said. This wouldn't add full color support to cmdline, though.

Issue #1256, opened 2021-01-31
[Feature request] Add copy button for website section link


Pieter12345:
# Feature request
Every now and then I link a MS docs page, and it would be super useful if I could click a section (or a button behind it) to copy a link to the section, rather than to the whole page. For example on https://methodscript.com/docs/3.3.4/Advanced_Guide.html, I'd like to be able to link the recompile section directly (this even works https://methodscript.com/docs/3.3.4/Advanced_Guide.html#recompile), but there's no button to copy to get that link directly.

# Related info
[Information on setting up a local website testing environment](https://methodscript.com/docs/3.3.4/SiteDeployTool.html#Testing%20Locally)
[Section header generation for the website](https://github.com/EngineHub/CommandHelper/blob/5e04ddf794d4e026a276db07e4c1868f996f3c52/src/main/resources/siteDeploy/resources/js/wiky.js#L27)

LadyCailin:
The code that runs the copyButton in the code blocks should be copied to make this work.

PR #1289, opened 2021-08-22
Addresses #1282

diaoright:


LadyCailin:
Hmm, doesn’t this logic need to go in the Static.MCColorToANSI function, rather than here? As far as I know, the in game colors were correct, and support any given RGB value, and don’t need to be perceptually scaled, however, the colors in a terminal do. Or maybe I’m misunderstanding what the PR does.

diaoright:
> Hmm, doesn’t this logic need to go in the Static.MCColorToANSI function, rather than here? As far as I know, the in game colors were correct, and support any given RGB value, and don’t need to be perceptually scaled, however, the colors in a terminal do. Or maybe I’m misunderstanding what the PR does.

Maybe I misunderstood what the purpose of fromRGBValue was?
Actually, MCToANSI merely does a replacing of the game codes to TermColors. Nonetheless, it still needs fromRGBValue output from color() and colorize().
The old method merely splitted the given hex code and making the last the one in effect.

LadyCailin:
Well, maybe there’s an additional bug I’m not aware of, but at least the point of #1282 was that using color(‘#123456’) in cmdline would end up with the color being stripped out. #1246 originally supported the RGB format in Minecraft, but since there’s a small finite set of valid colors for the terminal, they would have to be scaled or removed, and the first attempt was just to remove them. So that PR was merged in and the #1282 feature request created for cmdline. 

But if you’re suggesting that #1256 has a bug, and this fixes it, that’s great, but the title is misleading. In any case, I’ll have to take a deeper look at this later, when I’m not on my phone :D

PseudoKnight:
fromRGBValue converts hex `#FF0000` to the special Spigot supported format `§xFF0000`, where Spigot converts this later to a text component with an RGB color that MC can use. Changing this would break RGB colors in Minecraft.

Static.MCColorToANSI is called to convert legacy MC text formatting codes, like `§a` and now this special Spigot RGB format, to terminal formatting codes. 

I think perhaps some of this should be moved to MCColorToANSI, and some to TermColors, where a new method can be added to do the actual conversion of RGB to TermColor.

diaoright:
So it looks like I committed (PR'ed, at that) a mistake.

The output of printing to the terminal with color0 and/or colorize(), contrary to the stated at #1282, is not stripped of colors, but also accompanied by the §x[$...] pattern. My solution aimed to take advantage of MCColorToANSI also, which maps Minecraft (and conveniently Windows) to TermColor instances. However, a solution to the wrong problem in the wrong circumstances.

PR #1291, opened 2021-10-20
Warn on multiple block comment opens

LadyCailin:


LadyCailin:
On further thought, this is already a feature toggle, users can add the suppression to .msfileoptions, at the root, and get the suppression globally.

PR #1294, opened 2021-11-04
Generics take2

LadyCailin:


src/main/java/com/laytonsmith/core/functions/World.java
```diff
@@ -66,6 +67,7 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+@SuppressWarnings("ALL")
```
Pieter12345, line 70: I happened to be looking at this, and suppressing all warnings is generally a bad idea for future code quality.

LadyCailin: Oh.. I have no clue how this got added. It must have been me trying to suppress a single type of warning, but IntelliJ is StupidJ.

src/main/java/com/laytonsmith/core/constructs/InstanceofUtil.java
```diff
@@ -50,131 +72,329 @@
...
+				|| CClassType.AUTO.equals(instanceofThis) // either side
+				) {
 			return true;
 		}
-		if(type == null || type == CVoid.TYPE || instanceofThis == CVoid.TYPE || instanceofThis == CNull.TYPE) {
+		if(type == null // type is java null defined as false (except if instanceofThis was true, which is caught above)
+				|| CVoid.TYPE.equals(type) // void is not instanceof anything
+				|| CVoid.TYPE.equals(instanceofThis) // nothing is instanceof void
+				|| CNull.TYPE.equals(instanceofThis) // nothing is instanceof null (should be compile error)
+				|| CNull.TYPE.equals(type) // type is mscript null defined as false
```
Pieter12345, line 218: Ah, now I see why you've added the null check in `StaticAnalysis`. This method used to only be used by SA, and all instanceof checks there check whether a type is usable as some parameter/variable type, in which case `CNull` is suitable. But this is a fair change, as it actually isn't instance of anything. If you've added null checks in all places where SA uses this method, then this is all good.

Pieter12345 (commented):
This is only a partial review, as the changes in this PR are huge. I've read through at least SA and function signatures, which overall look good.

src/main/java/com/laytonsmith/core/constructs/InstanceofUtil.java
LadyCailin: There is still work to be done here, if you notice, this and the method below are nearly identical. They need to be merged and harmonized into a single implementation.

src/main/java/com/laytonsmith/core/events/Prefilters.java
```diff
@@ -80,8 +81,8 @@ private Prefilters() {
 	 */
 	@Deprecated
 	public static void match(Map<String, Mixed> map, String key,
-			String actualValue, PrefilterType type) throws PrefilterNonMatchException {
-		match(map, key, new CString(actualValue, Target.UNKNOWN), type);
+			String actualValue, PrefilterType type, Environment env) throws PrefilterNonMatchException {
```
PseudoKnight, line 84: Breaks extensions using old prefilters.

src/main/java/com/laytonsmith/core/natives/interfaces/Mixed.java
```diff
@@ -104,66 +109,59 @@
...
-	 * This method works with class type directly.
-	 *
-	 * Implementation note: The implementation of this should just be
-	 * {@link Construct#isInstanceof(com.laytonsmith.core.natives.interfaces.Mixed, java.lang.Class)} which supports
-	 * Mixed values.
-	 *
-	 * @param type
-	 * @return
-	 */
-	public boolean isInstanceOf(Class<? extends Mixed> type);
```
PseudoKnight, line 137: Breaks a few extensions, though some still use plain old instanceof.

src/main/java/com/laytonsmith/core/events/Event.java
```diff
@@ -59,7 +59,7 @@
 	 * Error.
 	 */
 	@Deprecated
-	public boolean matches(Map<String, Mixed> prefilter, BindableEvent e) throws PrefilterNonMatchException;
+	public boolean matches(Map<String, Mixed> prefilter, BindableEvent e, Environment env) throws PrefilterNonMatchException;
```
PseudoKnight, line 62: matches(), evaluate(), modifyEvent() (and maybe event convert()?) changes all break extensions.

src/main/java/com/laytonsmith/core/Static.java
```diff
@@ -795,8 +550,8 @@ public static UUID GetUUID(Mixed subject, Target t) {
 		return GetUUID(subject.val(), t);
 	}
 
-	public static MCOfflinePlayer GetUser(Mixed search, Target t) {
-		return GetUser(search.val(), t);
+	public static MCOfflinePlayer GetUser(Mixed search, Target t, Environment env) {
```
PseudoKnight, line 553: These two GetUser() methods are probably used by extensions, but I can't think of one off the top of my head.

PseudoKnight (commented):
Here are some breaking changes I noticed. It's harder to do an exhaustive sweep of extensions because aggressive deprecations don't differentiate between methods that exist and don't. We're probably going to need a version release prior to pulling this. That way extensions can continue to build. We should probably also make sure the final build installs in Jenkins. It's currently our only repo for anything later than early 3.3.4 snapshots.

src/main/java/com/laytonsmith/core/constructs/CArray.java
```diff
@@ -567,7 +726,7 @@ protected String getString(Stack<CArray> arrays, Target t) {
 	}
 
 	@Override
-	public long size() {
+	public long size(Environment env) {
```
PseudoKnight, line 729: Another breaking change.

src/main/java/com/laytonsmith/core/natives/interfaces/Mixed.java
LadyCailin: As discussed, extensions can use InstanceofUtil.isInstanceof (or just java instanceof), which exists today, and will continue to exist. This will need to be updated now, but can be done immediately.

src/main/java/com/laytonsmith/core/events/Event.java
LadyCailin: I'm skipping convert, but everything else is using reflection to recall the method now.

src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCCommand.java
```diff
@@ -209,26 +213,33 @@ public String toString() {
 
 	@Override
 	public List<String> handleTabComplete(MCCommandSender sender, String alias, String[] args) {
+		Environment env;
+		try {
+			env = Static.GenerateStandaloneEnvironment();
```
LadyCailin, line 218: I cannot imagine that this is right. Though, since we can't change the signature of this function, the solution may be pretty interesting as well.

Pieter12345: Can we not use the env from the tabcomplete closure below instead of creating a new env?

PseudoKnight: Grab the env from the closure?

PR #1295, opened 2021-11-11
Improve MAVEN build Performance

hj987654321:
Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -483,6 +483,8 @@
 					</compilerArgs>
 					<fork>true</fork>
 					<optimize>true</optimize>
+							<useIncrementalCompilation>true</useIncrementalCompilation>
+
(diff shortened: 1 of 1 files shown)


LadyCailin:
Hi, thanks for your submission, but I'm going to opt to not merge this. Most of the compilation time comes from shading and custom compile checks, not the actual javac process, and introducing this saves a few second in the build, but introduces more opportunities to need to do a clean and build anyways. Additionally, the spacing is wrong and master is currently under a code freeze.

PR #1300, opened 2022-02-24
Allow command execution from RCON

addstar:
This fixes an issue where CommandHelper aliases cannot be executed via RCON. Attempting to run a command results in nothing happen, there is no response and the command doesn't execute.

With this change commands are properly executed and command output is sent to the RCON client as expected. I have used this extensively for a few weeks and I've experienced no issues.

### DIFF
--- a/src/main/java/com/laytonsmith/commandhelper/CommandHelperPlugin.java
+++ b/src/main/java/com/laytonsmith/commandhelper/CommandHelperPlugin.java
@@ -71,6 +71,7 @@ import org.bukkit.command.BlockCommandSender;
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.ConsoleCommandSender;
+import org.bukkit.command.RemoteConsoleCommandSender;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
 import org.bukkit.event.EventException;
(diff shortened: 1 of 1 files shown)


PseudoKnight:
This makes /runalias work for RCON, but not aliases directly. This is not a problem with the PR, but functionality could be expanded. 

We might need to consider the command sender name for RemoteConsoleCommandSender.

LadyCailin:
Do multiple users have individual access to RCON, or is it all the same user/password combination?

addstar:
> This makes /runalias work for RCON, but not aliases directly.
Oh sorry! Yes I forgot to mention this part. I'm not sure what would be involved in making this work, but it wasn't the functionality I needed :slightly_smiling_face: There was literally no way to call a CH alias from RCON so I dug in to the reason and fixed it.

> Do multiple users have individual access to RCON, or is it all the same user/password combination?
No, there's only a single password for RCON access and no user/name: https://wiki.vg/RCON

LadyCailin:
Given that, keeping ~console is fine.

PseudoKnight:
Keeping? As far as I know, player() will not return `~console` for RemoteConsoleCommandSender.

addstar:
> Keeping? As far as I know, player() will not return `~console` for RemoteConsoleCommandSender.

Hmm.. good point. I didn't test the `player()` result.. but a quick search in the code makes me think you're right. I think this might also need to be changed for it to work properly: [src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCCommandSender.java](https://github.com/EngineHub/CommandHelper/blob/master/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCCommandSender.java#L66)

PseudoKnight:
It actually returns as `Rcon`. Though looking more into rcon I wouldn't recommend it.

PseudoKnight:
Keep in mind that it doesn't even encrypt the rcon password when its sent.

LadyCailin:
Fixed in: https://github.com/EngineHub/CommandHelper/commit/78e9f3b853eda64f92ff9cfafd8e5139d5702179

And yeah, RCON is probably a bad idea anyways, but it's not us opening that can of worms, so I'm fine with supporting it at a very basic level.

addstar:
Completely understood :+1: We're only using it via localhost (from automation) and the system is protected, so the security of rcon isn't really a concern for us.

addstar:
Side note: Thanks for being so active and accepting the change :heart: We've been using CH extensively for many years and love the power it can provide! Couldn't run our servers without it :100:

PseudoKnight:
I added direct alias support too, for good measure.

PR #1301, opened 2022-03-04
Add first version of deep bind compilation checks.

LadyCailin:
All events will need to be rewritten to fully support this, and from a
framework perspective, there are still missing items (particularly
around generation of the docs) but what is here so far does work.

In general, this changes the format of the prefilters from a procedural
to a declarative format. This makes specifying the event a bit more
cumbersome, but at a huge payoff. For one, the docs and filtering are
harder to screw up, but primarily, this should be a substantially better
performing pattern. Previously, many of the events used a
PrefilterNonMatchException to signal that the prefilter doesn't match.
This was less code, but throwing exceptions is substantially more
expensive than simply returning a false boolean. However, doing a
procedural approach and only dealing with booleans would be even more
code than this new approach is, so this is a good balance between
performance and development effort.

Remaining work to be done is to obviously convert every event, but this
can be done at a slower pace in the future. For the framework, the
remaining item is to add a way to provide the prefilter description in
the event declaration, and then using that in the documentation
generation as well.

The compilation works, but the new mechanism for checking the prefilters
at runtime has not been tested, and this PR should not be merged in
until it has.


src/main/java/com/laytonsmith/core/ParseTree.java
```diff
@@ -389,4 +393,29 @@ public Target getTarget() {
...
+			return ivar.getDefinedType();
+		} else if(getData() instanceof CFunction cf) {
+			List<CClassType> argTypes = new ArrayList<>();
+			List<Target> argTargets = new ArrayList<>();
+			Set<ConfigCompileException> exceptions = new HashSet<>();
+			for(ParseTree child : getChildren()) {
+				argTypes.add(child.getType(env));
+				argTargets.add(child.getTarget());
+			}
+			return cf.getCachedFunction().getReturnType(getTarget(), argTypes, argTargets, env, exceptions);
```
PseudoKnight, line 415: This doesn't do anything with possible exceptions. Not sure if this is intended to be added later.

src/main/java/com/laytonsmith/core/events/EventUtils.java
```diff
@@ -181,12 +181,105 @@ public static void ManualTrigger(String eventName, CArray object, Target t, bool
...
+					// The compiler should have already warned about this
+					continue;
+				}
+				Pair<Prefilters.PrefilterType, Event.PrefilterMatcher> pair = prefilters.get(prefilter.getKey());
+				Event.PrefilterMatcher matcher = pair.getValue();
+				Mixed value = prefilter.getValue();
+				switch(pair.getKey()) {
+					case STRING_MATCH: {
+						String compare = matcher.getStringParameter(e);
+						if(compare != null && !Prefilters.FastStringMatch(compare, value.val())) {
```
PseudoKnight, line 215: This means there's no way to handle case sensitivity in event prefilters using this new method. STRING_IC_MATCH?

PseudoKnight: There's also no way to handle nulls.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -910,6 +910,67 @@ public boolean matches(Map<String, Mixed> prefilter, BindableEvent e) throws Pre
...
+								return pie.getItem().getType().getName();
+							}
+							return null;
+						}
+					})
+					.set("block", PrefilterType.STRING_MATCH, new PrefilterMatcher() {
+						@Override
+						public String getStringParameter(BindableEvent e) {
+							if(e instanceof MCPlayerInteractEvent pie) {
+								return pie.getClickedBlock().getType().getName();
```
PseudoKnight, line 945: This can be null. For string matches we can return the String "null" in that case, unless you have a better idea.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -910,6 +910,67 @@ public boolean matches(Map<String, Mixed> prefilter, BindableEvent e) throws Pre
...
+								}
+							}
+							return null;
+						}
+					})
+					.set("itemname", PrefilterType.STRING_MATCH, new PrefilterMatcher() {
+						@Override
+						public String getStringParameter(BindableEvent e) {
+							if(e instanceof MCPlayerInteractEvent pie) {
+								return pie.getItem().getType().getName();
```
PseudoKnight, line 936: Item can be null in addition to "block" type.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -910,6 +910,67 @@ public boolean matches(Map<String, Mixed> prefilter, BindableEvent e) throws Pre
 			return false;
 		}
 
+		@Override
+		public PrefilterBuilder getPrefilters() {
+			return PrefilterBuilder
+					.start("button", PrefilterType.STRING_MATCH, new PrefilterMatcher() {
```
PseudoKnight, line 916: This is an example of a prefilter where case is current insensitive.

src/main/java/com/laytonsmith/core/functions/EventBinding.java
```diff
@@ -233,16 +246,122 @@ public boolean useSpecialExec() {
...
+											} else if(value.isConst()) {
+												try {
+													Pattern.compile(value.getData().val());
+												} catch (PatternSyntaxException ex) {
+													exceptions.add(new ConfigCompileException(ex.getMessage(), t));
+												}
+											}
+											break;
+										case STRING_MATCH:
+											if(!valType.unsafeDoesExtend(CString.TYPE)) {
```
PseudoKnight, line 335: Gets "ClassNotFoundException: auto is not a real type" when using a function with a return type of auto.

src/main/java/com/laytonsmith/core/ParseTree.java
LadyCailin: Hmm, that's a good point. In general, this shouldn't throw an exception, because of the semi-declarative nature, but for malformed functions, that would themselves be giving out compile errors, that's a good point, and I'm not sure what the behavior would be.

src/main/java/com/laytonsmith/core/events/EventUtils.java
LadyCailin: Hmm, don't think that was possible already, I'm pretty sure I kept the same behavior as before. https://github.com/EngineHub/CommandHelper/blob/master/src/main/java/com/laytonsmith/core/events/Prefilters.java#L169 Or are you suggesting a new feature?

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
LadyCailin: Oh I see. Yeah, that's just because the value is lowercased. Ok, good catch, I'll have to give this some thinks.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
LadyCailin: Need to give this some thinks too.

PseudoKnight: At least for String matches it should be as easy as returning a String "null". However, other match types might not be so fortunate. I'm not sure if there are any others that support null that aren't strings.

src/main/java/com/laytonsmith/core/ParseTree.java
LadyCailin: The function signature doesn't currently allow for exceptions to be thrown, so I'm just going to leave this as won't fix for now. However, I suspect that this will crop up again later, but it will be easier to understand the expected behavior then, and the solution here might be more obvious.

src/main/java/com/laytonsmith/core/events/drivers/PlayerEvents.java
```diff
@@ -857,57 +888,53 @@ public void bind(BoundEvent event) {
...
-					MCMaterial mat = pie.getItem().getType();
-					if(mat == null) {
-						if(!(item instanceof CNull)) {
-							return false;
+					})
+					.set("block", "The block type the player interacts with, or null if nothing",
+							new MaterialPrefilterMatcher<>() {
+						@Override
+						public MCMaterial getMaterial(MCPlayerInteractEvent pie) {
+							return pie.getClickedBlock().getType();
```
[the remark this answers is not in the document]
LadyCailin: I'm tempted to make these subclasses for MCItemStack and MCBlock, so the null handling logic can live in only one place.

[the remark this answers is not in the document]
PseudoKnight: I'm not exactly sure what you mean, but keep in mind that whether an event can return null for an item or block (etc) is specific to the event. So writing it in the event's prefilter should be fine.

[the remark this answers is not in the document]
LadyCailin: I mean that if the block or item stack is null, then the material would also be null, which would then be compared against CNull. The check would look the same - if the method would never return null, then the if(block == null) { return null; } would never happen anyways, but in the events where it can happen, it wouldn't NPE, it would just return null.

src/main/java/com/laytonsmith/core/events/EventUtils.java
```diff
@@ -181,12 +182,62 @@ public static void ManualTrigger(String eventName, CArray object, Target t, bool
...
+			PrefilterBuilder prefilterBuilder = driver.getPrefilters();
+			if(prefilterBuilder != null) {
+				prefilters = prefilterBuilder.build();
+			} else {
+				prefilters = null;
+			}
+			prefilterData.put(driver, prefilters);
+		} else {
+			prefilters = prefilterData.get(driver);
+		}
```
PseudoKnight, line 208: Could each event's built prefilters be stored in AbstractEvent instead?

LadyCailin: Hmm, not without changing the Event interface again, or rather, changing all the subclasses. It's also ok for the cache to live outside the Event objects, so we don't violate the single responsibility principal too badly.

PseudoKnight: What's wrong with changing the Event interface is if it's implemented in AbstractEvent?

LadyCailin: Alright, that cleaned things up a lot nicer actually, so good call.

PR #1304, opened 2022-03-11
Reload rewrite

### LINKED ISSUE #1156: Uncaught exception handler is not shared
CMDHELPER-3162 - Reported by PseudoKnight

When setting the uncaught exception handlers, aliases do not use it because they create an entirely new environment every time. 

https://github.com/EngineHub/CommandHelper/pull/294
https://github.com/EngineHub/CommandHelper/pull/458

Neither solution was correct. I'm thinking for now we need to have a new environment created on reload, clone that for main.ms, also store it in aliascore so that any time an alias is run it can clone it. That way any uncaught exception handlers defined in main.ms (or other scripts) can be referenced in aliases. Also, unlike my previous PR, this would not affect future systems that add separate distinct environments. 

I would also like to discuss the possibility of not executing auto_includes every time an alias is run. Unless there's some dynamic code in an auto_include, which I would think would be bad practice, it wouldn't affect behavior if we just used the procs we got when we execute on reload. This would even further reduce alias run overhead, but would need to pass it by LC.

### LINKED ISSUE #1306: Add "halt-on-failure" option
Some servers may have some security code written in MethodScript, which, if there was some misconfiguration or compile error, and the system couldn't start up properly, would rather keep the server shut down, rather that up without the protections. Therefore, expanding the halt-on-failure option to shut down the server if on startup it cannot start in a consistent state would make sense for these servers. Especially after the reload rewrite, we can keep the behavior as-is today for second and on recompiles, (that is, print the errors, but keep the old, known working configuration running), but then on startup, when there is no "known working configuration", simply shutting down the server instead.

PseudoKnight:
This is a step towards async compilation support as well as Static Analysis compatibility with LangServ/vscode.

- Separates compiling and execution so that reloading can be stopped on compile errors
- Shutdown server on first load when halt-on-failure is true (Closes #1306)
- Decouples AliasCore from CommandHelperPlugin
- Moves LocalPackages out of AliasCore
- Now reloads Preferences
- Makes IncludeCache non-static and stores it in StaticRuntimeEnv
- Moves autoIncludeAnalysis to StaticRuntimeEnv
- Moves ExecutionQueue to StaticRuntimeEnv
- Moves uncaughtExceptionHandler to StaticRuntimeEnv (Closes #1156)
- Fixes ExecutionQueue so that it properly stops


src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -155,25 +140,26 @@ public boolean alias(String command, final MCCommandSender sender) {
...
-				parent.persistenceNetwork, parent.profiles, new TaskManagerImpl());
-		CommandHelperEnvironment cEnv = new CommandHelperEnvironment();
-		cEnv.SetCommandSender(sender);
-		cEnv.SetCommand(command);
-		Environment env = Environment.createEnvironment(gEnv, staticRuntimeEnv, cEnv, compilerEnv);
+		Environment env;
+		try {
+			env = environment.clone();
+		} catch (CloneNotSupportedException ex) {
+			throw new Error(ex.getMessage());
```
Pieter12345, line 147: Assuming we don't expect this case to ever happen, it'd be best to throw `new Error(ex)` instead, passing the original stacktrace in the `caused by` exception stacktrace.

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -215,127 +201,123 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
...
-				}
-			}
-			if(options.reloadTimeouts()) {
-				ProfilePoint clearingTimeouts = parent.profiler.start("Clearing timeouts/intervals", LogLevel.VERBOSE);
-				try {
-					Scheduling.ClearScheduledRunners();
-				} finally {
-					clearingTimeouts.stop();
-				}
+				env = newEnv.clone();
```
Pieter12345, line 296: I'm not understanding why `newEnv` is cloned here, right after it has been created. Why is `newEnv` itself used as the environment, but `env` is used for the whole compilation and setup process of the MS code?

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -344,121 +326,218 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
...
+			ProfilePoint compilerMS = profiler.start("Compilation of MS files in Local Packages", LogLevel.VERBOSE);
 
-				// Unregister previously registered commands
-				{
-					MCCommandMap map = Static.getServer().getCommandMap();
-					for(String command : registeredCommands) {
-						map.getCommand(command).unregister(map);
+			// Set and analyze auto includes for static analysis.
+			Set<ConfigCompileException> compileExceptions = new HashSet<>();
+			StaticAnalysis.setAndAnalyzeAutoIncludes(
```
Pieter12345, line 357: This is still static, which forms an issue when 'reverting back' (or rather not touching) to the old execution environment. If in that execution environment a call is made to `eval()`, or something is dynamically `include()`d, it will use the newly set auto includes to perform static analysis on. I suppose that `StaticAnalysis.autoIncludesAnalysis` could be moved to the `StaticRuntimeEnv` environment instead (just like the `IncludeCache`), resulting in having one shared `autoIncludesAnalysis` per recompile. This does not affect LangServ for now, as it has no notion of auto includes.

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -344,121 +326,218 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
...
+		ProfilePoint postReloadAliases = profiler.start("Extension manager post reload aliases", LogLevel.VERBOSE);
 		try {
 			ExtensionManager.PostReloadAliases();
 		} finally {
 			postReloadAliases.stop();
 		}
+
+		String output;
+		if(firstLoad) {
+			output = "Loading complete. ";
```
Pieter12345, line 529: This is a bit picky, but I'd personally prefer either `Load` and `Reload` or `Loading` and `Reloading` in these two cases of feedback.

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -678,8 +757,7 @@ public static String file_get_contents(String fileLocation) throws IOException {
 	public static boolean file_put_contents(File fileLocation, String contents, String mode)
 			throws Exception {
 		BufferedWriter out = null;
```
Pieter12345, line 759: The `null` value here is not used, and you've correctly removed the handling of the never-happening `null` value. You should however also remove the `null` assign here, as it is never used.

src/main/java/com/laytonsmith/core/functions/IncludeCache.java
```diff
@@ -52,19 +62,24 @@ public static ParseTree get(File file, com.laytonsmith.core.environments.Environ
 	public static ParseTree get(File file, com.laytonsmith.core.environments.Environment env,
```
Pieter12345, line 62: I think it would be cleaner if a non-static `get()` method was added for both of these static getters, having the static versions call the non-static versions on the `IncludeCache` from their environment. This will not break extensions that are using this, and it allows for calling these getters directly on the `IncludeCache` instance whenever that's convenient.

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -344,121 +326,218 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
 				}
 			}
 
-			if(!Prefs.isInitialized()) {
-				Prefs.init(prefFile);
+			// Get the default files (main.ms, aliases.msa, auto_include.ms)
+			try {
+				String main = file_get_contents(mainFile.getAbsolutePath());
```
LadyCailin, line 331: We should probably remove these methods, and use the normal utility methods. What is this, [PHP](https://www.php.net/manual/en/function.file-get-contents.php)?!?

src/main/java/com/laytonsmith/core/AliasCore.java
PseudoKnight: This stores the new environment (if compilation is successful) in AliasCore for usage in alias() and elsewhere, though StaticRuntimeEnv is the important bit.  Before it would create a new StaticRuntimeEnv from fields stored in CommandHelperPlugin. So, I'm cloning here to better ensure the compilation and execution of the ms files doesn't modify the environment. Maybe we only need to store StaticRuntimeEnv in AliasCore, though. But there are some things that arguably should be moved from GlobalEnv to StaticRuntimeEnv but haven't yet, like the previously mentioned execution queue and the uncaughtExceptionHandler.

src/main/java/com/laytonsmith/core/functions/IncludeCache.java
PseudoKnight: I thought about that. In practice it doesn't seem to make things that much cleaner. This method is used so infrequently anyway, so I didn't think it was worth changing.

src/main/java/com/laytonsmith/core/AliasCore.java
PseudoKnight: I'm thinking moving the execution queue and uncaught exception handler to StaticRuntimeEnv is the appropriate move here. Are there others that need to be moved too? Then we can just generate a new GlobalEnv for aliases and avoid cloning.

Pieter12345: I believe that the uncaught exception handler currently works per compilation unit (alias, main ms file). This might have been changed in the past though, but it would be good to check that. Moving it to StaticRuntimeEnv would share it across all compilation units. I agree that the execution queue would fit the StaticRuntimeEnv.

Regarding the cloned environment: I think that if the compiler makes changes to an environment that is not going to be used, that we are looking at a bug there. Especially if this is just about the uncaught exception handler and the execution queue, which both cannot be set/changed/used in compile time as far as I am aware.

Pieter12345:
> I could make this static again and just pull the autoIncludeAnalysis from the env. Would that work for now?
That would work indeed. You have a good point that everything is final there, and actually so should the autoIncludeAnalysis be (switching this halfway compilation is not supported). So an `IllegalStateException` could be thrown when the `setAndAnalyzeAutoIncludes` method is called for a second time (the first time, the `endScope` field will be `null`, that check would be sufficient). This essentially makes it final, and causes an exception when we at some point make a mistake in not cloning an environment or setting auto includes in a wrong environment.

src/main/java/com/laytonsmith/core/AliasCore.java
PseudoKnight: May be useful to refer to #1156. It even links a PR where you've actually tried to fix this issue before.

Pieter12345: Oh wow, that was 7 years ago, I doubt that I knew what I was doing exactly. But we both had the idea that the uncaught exception handler should be shared, so maybe this is the time to finally do that by doing what you proposed. There was no StaticRuntimeEnv back then, so the solution was not obvious before. Maybe @LadyCailin wants to share an opinion on making the uncaught exception handler shared among all compilation units as well?

LadyCailin: Yes, moving this into the StaticRuntimeEnv makes sense, now that that is a thing.

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -215,127 +199,123 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
...
+			extensionPreReload.stop();
+		}
 
-			ProfilePoint shutdownHooks = parent.profiler.start("Shutdown hooks call", LogLevel.VERBOSE);
+		Profiles profiles;
+		try {
+			profiles = new ProfilesImpl(fileLocations.getProfilesFile());
+		} catch (IOException | Profiles.InvalidProfileException ex) {
+			MSLog.GetLogger().e(MSLog.Tags.GENERAL, ex.getMessage(), Target.UNKNOWN);
+			return;
```
LadyCailin, line 247: Hmm, what will this do on first load? Just prevent CH from loading up? Would it make sense to do something else here?

src/main/java/com/laytonsmith/core/AliasCore.java
```diff
@@ -215,127 +199,123 @@ public final void reload(MCPlayer player, String[] settings, boolean firstLoad)
...
-				StaticLayer.GetConvertor().runShutdownHooks();
+				DataSourceFactory.DisconnectAll();
+				ConnectionMixinFactory.ConnectionMixinOptions opts = new ConnectionMixinFactory.ConnectionMixinOptions();
+				opts.setWorkingDirectory(fileLocations.getConfigDirectory());
+				persistenceNetwork = new PersistenceNetworkImpl(fileLocations.getPersistenceConfig(),
+						new URI("sqlite://" + fileLocations.getDefaultPersistenceDBFile().getCanonicalFile().toURI()
+								.getRawSchemeSpecificPart().replace('\\', '/')), opts);
+			} catch (DataSourceException | URISyntaxException | IOException ex) {
+				Static.getLogger().log(Level.SEVERE, null, ex);
+				return;
```
LadyCailin, line 262: Same here. Maybe we ought to have an option to "stop server startup on failure", since one might rely on CH for some security, and would rather the server not start up if there's a misconfiguration, rather than startup without CH.

PseudoKnight: Ya, it's a tough one. We have a halt-on-failure preferences option. We could make it safely shutdown the server if it's on firstload, keep the old profiles/etc if not on reload.

LadyCailin (approved):
Other than the one thing I suggested (and you're up for doing it now) this looks good!

PseudoKnight:
There's still some more related things that could be done, especially if we're going to make this async, but I think this is good for now.

PR #1307, opened 2022-03-18
Let x_find accept 3 arguments, as intended.

Lildirt:
tested using 3 args, seems to work fine.

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/Cmdline.java
+++ b/src/main/java/com/laytonsmith/core/functions/Cmdline.java
@@ -2008,7 +2008,7 @@ public class Cmdline {
 
 		@Override
 		public Integer[] numArgs() {
-			return new Integer[]{1, 2};
+			return new Integer[]{1, 2, 3};
 		}
 
 		@MEnum("ms.lang.FindType")


Pieter12345:
While we are at it, is this documentation correct? `array<string> {regex, [startFrom], [type] | regex, [startFrom], types}`
Looking at the code, I think that the actual type is `array<string> {regex, [startFrom], [types]` instead, where `types` is an array (read through `ArgumentValidation.getEnumSet`).

LadyCailin:
Hmm, I think it's correct? We have 6 allowed permutations:
1. regex
2. regex, type
3. regex, types
4. regex, startFrom
5. regex, startFrom, type
6. regex, startFrom, types

Expanding out the signature permutations, we get:
1. (sig 1) regex
2. (sig 1) regex, startFrom
3. (sig 1) regex, type
4. (sig 1) regex, startFrom, type
5. (sig 2) regex, types
6. (sig 2) regex, startFrom, types

So looks like it lines up to me?

I'm going to go ahead and merge this, since this is definitely wrong, but if there's something I'm missing, we can always iterate more.

PR #1309, opened 2022-03-24
Throw compile errors in strict mode for illegal statements

PseudoKnight:


LadyCailin (approved):

src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
```diff
@@ -2216,6 +2216,27 @@ private static ParseTree postParseRewrite(ParseTree ast, Environment env,
...
 				ast.getChildren().set(i, newChild);
 				i--; // Allow the new child to do a rewrite step as well.
+				continue;
+			}
+
+			// In strict mode throw compile errors when encountering child statements in function arguments where
+			// statements are not acceptable because void would be an invalid argument type. This would otherwise be
+			// a runtime error in strict mode where auto-concat is not allowed and statements are used instead.
+			// This can be removed once a more comprehensive void return type check is implemented.
+			if(ast.getFileOptions().isStrict() && child.getData().val().equals(Compiler.__statements__.NAME)) {
```
Pieter12345, line 2226: This should contain an instanceof CFunction to not match `__statements__` string literals and other terms that stringify to this.

src/main/java/com/laytonsmith/core/MethodScriptCompiler.java
```diff
@@ -2216,6 +2216,27 @@ private static ParseTree postParseRewrite(ParseTree ast, Environment env,
...
+				node = ast.getData();
+				if(!(node instanceof CFunction)) {
+					continue;
+				}
+				Function function = ((CFunction) node).getCachedFunction();
+				if(function instanceof BranchStatement) {
```
Pieter12345, line 2232: You could use a new Java feature here!
```java
if(function instanceof BranchStatement branchStatement) {
    branchStatement.isBranch(...);
}
```

Pieter12345 (commented):
Looks good, small change requested.

PseudoKnight:
Okay, that makes things very clean. Much simpler than my previous check.

PR #1310, opened 2022-03-25
[WIP] Function signature implementation

Pieter12345:
The purpose of this PR is to add function signature implementations to existing functions, replacing `getReturnType()` overrides where possible. A few function signatures will be added to this PR every now and then, and they could be merged every now and then as well.


LadyCailin:
Base this commit off of the genericsTake2 branch. I kept making signature changes to typecheck and getReturnType, and I got tired of changing that everywhere, so I just made those functions all use function signatures now, so this might be mostly done already.

PR #1312, opened 2022-04-02
Process keywords faster

PseudoKnight:
This avoids traversing children twice when keywords are found, and removes a redundant keyword hashmap lookup.

processKeywords easily takes the majority of compile time on my code.


PseudoKnight:
Wait, this could be better.

LadyCailin (approved):

PR #1313, opened 2022-04-05
Display exceptions on auto includes in LangServ

Pieter12345:


src/main/java/com/laytonsmith/core/functions/IncludeCache.java
```diff
@@ -135,20 +153,38 @@ public static ParseTree get(File file, com.laytonsmith.core.environments.Environ
...
-			String fileName = (ex.getFile() == null ? "Unknown Source" : ex.getFile().getName());
-			throw new CREIncludeException("There was a compile error when trying to include the script at " + file
-					+ "\n" + ex.getMessage() + " :: " + fileName + ":" + ex.getLineNum(), t);
-		} catch (ConfigCompileGroupException exs) {
-			StringBuilder b = new StringBuilder();
-			b.append("There were compile errors when trying to include the script at ").append(file).append("\n");
-			for(ConfigCompileException ex : exs.getList()) {
+			if(exceptions != null) {
+				exceptions.add(ex);
+				exceptions.add(new ConfigCompileException("Compile error(s) in included file at " + file, t));
```
LadyCailin, line 158: Can we do a compiler warning (or maybe even less) here? We can write a compiler warning into the environment, create a new suppression category, set it to medium, and it'll show up as a hint in the IDE, rather than an error. Every file that includes this file will also have this error, and so the actual problem will be drowned out by the red herrings, which is not intuitive. True, variables that are defined in the include will be broken in this file, but honestly, I don't want to cater to that style, especially if we're going to remove that ability later anyways.

Even if this were a normal compiler warning (and not an error) I'd support that, but this is just too much. For compilation where we expect the code to immediately run after, maybe this is ok, but during development when you're actively typing in an auto_include file, and you're not done typing, but suddenly there are 75 errors in your window.. that's obnoxious.

LadyCailin: No, I don't think I support a compiler warning either, actually. I think a hint is the most I'd support, and even then, I'm ok with just simply removing this entirely.

Pieter12345: So in runtime it makes sense to include which `include()` call caused the problem, but in compile time with an IDE this indeed makes less sense. Keeping runtime behavior the same, I'll remove the exception(s) on `include(...)` in compile time. We can always re-add something here if this causes undesired scenarios.

LadyCailin (approved):

PR #1322, opened 2022-07-03
Minor fixes

aah-uuh:
Some trivial fixes. Most just contemplate abrupt terminations on the interpreter.


LadyCailin:
Thanks! These are good little fixes.

PR #1324, opened 2022-08-24
Fix Event API Link

RitzKid76:
The link caused a 404 while it should have gone to the Event API page


PseudoKnight:
It should be `Event_API`.

PR #1347, opened 2023-01-13
Fix self statement semicolon rewrite

Pieter12345:
- Fix semicolons not getting added to self statements where placement was intended.
- Add documentation.

While this fixes the intended behavior of the `addSelfStatements` method, two unit tests are broken since this newly added `__autoconcat__` ends up in the runtime AST as `__statements__`:
```
[ERROR]   OptimizationTest.testSwitchInSwitch:410 expected:
<...s__(switch(dyn(1),2,[switch(dyn(4),5,__statements__(msg('hi')),6,__statements__(msg('hi']))),3,__statements__...>
but was:
<...s__(switch(dyn(1),2,[__statements__(switch(dyn(4),5,__statements__(msg('hi')),6,__statements__(msg('hi')]))),3,__statements__...>

[ERROR]   OptimizationTest.testUnreachableCodeComplex:164 expected:
<...(msg('a'),if(dyn(1),[ifelse(dyn(1),__statements__(die()),dyn(2),__statements__(die()),__statements__(die(]))),__statements__(m...>
but was:
<...(msg('a'),if(dyn(1),[__statements__(ifelse(dyn(1),__statements__(die()),dyn(2),__statements__(die()),__statements__(die()]))),__statements__(m...>
```

@LadyCailin What is the intended behavior here?


LadyCailin:
Intended behavior is that switch with inner statements is also a statement. Same for ifelse. It's odd that the outer one isn't a self-statement as well, I suppose it's likely because of the order of operations in the tree descent? Anyways, I would expect the outer switch/if to be in a statement as well, but the fact that the inner ones are statements is expected.

PR #1366, opened 2023-09-16
Add support for custom tags in item meta

PseudoKnight:
This is a proposal solution for adding custom tags from Bukkit's PersistentDataContainers in item meta. This is for preserving tags added by other plugins as well as allowing scripters more control over their own custom data stored in items. 

I'm using array objects to store NBT types along with the values. Due to API limitations we're not able to tell what type of data is under each key, so all types are checked, ordered by an estimation of frequency of use. This can be optimized later, but performance actually seemed adequate and doesn't rely on reflection.

API naming worth discussing is "type", "value", and "tags". We could also go for something like "custom" instead of "tags", "data" instead of "value"

Item array example:
```
{
	"name": "IRON_PICKAXE",
	"meta": {
		"tags": {
			"commandhelper:mytag": {
				"type": "STRING",
				"value": "something or other",
			},
			"mcmmo:super_ability_boosted": {
				"type": "INTEGER",
				"value": 0,
			}
		}
	}
}
```
Explosive pick example:
```
bind('block_break', null, null, @event) {
	@item = pinv(player(), null);
	if(@item 
	&& string_ends_with(@item['name'], 'PICKAXE')
	&& @item['meta'] 
	&& @item['meta']['tags'] 
	&& array_index_exists(@item['meta']['tags'], 'commandhelper:explosive_pick')) {
		@size = @item['meta']['tags']['commandhelper:explosive_pick']['value'];
		explosion(@event['location'], @size, false, false, puuid());
	}
}
```
Creating explosive pick:
```
pgive_item(array(
	name: 'DIAMOND_PICKAXE',
	meta: array(
		tags: array(
			'explosive_pick': array( // commandhelper namespace is added by default
				type: 'BYTE', 
				value: 1
			)
		)
	)
))
```

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/MCTagContainer.java
+++ b/src/main/java/com/laytonsmith/abstraction/MCTagContainer.java
@@ -2,7 +2,7 @@ package com.laytonsmith.abstraction;
 
 import com.laytonsmith.abstraction.enums.MCTagType;
 
-import java.util.Collection;
+import java.util.Set;
 
 /**
  * Minecraft NBT containers that can be used to read and modify tags in supported game objects.
@@ -17,12 +17,10 @@ public interface MCTagContainer extends AbstractionObject {
 	boolean isEmpty();
 
 	/**
-	 * Gets a set of key objects for each tag that exists in this container.
-	 * These are minecraft formatted namespaced keys. (e.g. "namespace:key")
-	 * These key objects can be passed to other methods in this class.
+	 * Gets a set of namespaced keys for each tag that exists in this container. (e.g. "namespace:key")
 	 * @return a set of keys
 	 */
-	Collection getKeys();
+	Set<MCNamespacedKey> getKeys();
 
 	/**
 	 * Returns the tag type with the given key.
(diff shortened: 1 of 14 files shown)


src/main/java/com/laytonsmith/abstraction/MCTagContainer.java
```diff
@@ -0,0 +1,67 @@
...
+	 */
+	boolean isEmpty();
+
+	/**
+	 * Gets a set of key objects for each tag that exists in this container.
+	 * These are minecraft formatted namespaced keys. (e.g. "namespace:key")
+	 * These key objects can be passed to other methods in this class.
+	 * @return a set of keys
+	 */
+	Collection getKeys();
```
LadyCailin, line 25: Collection of what? I guess object, but would be good to be explicit about it, `Collection<Object>`. Also, I'm starting to think that it would be better to have a wrapper class here, even if internally it just has the one Object field. That way, it's at least encapsulated to make it less error prone at the top level.

PseudoKnight: I was torn on this, but thought I'd see if it made sense to you, since we only ever use the string representation. (it's a string key that bukkit converts into an object that splits the namespace and key and then we convert back to a string, which felt wasteful) Anyway, updated.

LadyCailin: Yeah, preventing boxing feels like premature optimization. The additional clarity from ensuring correctness seems like it's worth more in this case.

LadyCailin (approved):

PR #1374, opened 2023-12-17
Add string_compare function and support integer return type in array_sort function

Pieter12345:
- Export functionality of Java's `String.compareTo(String)` in MethodScript API.
- Support integer return type in array_sort closure.

This is useful to convert hacky code such as:
```java
array @arr = array(array('key': 'Hello'), array('key': 'World'), array('key': 'Aaahhh'));
@arr = array_sort(@arr, closure(@left, @right) {
    return(array_sort(array(@left['key'], @right['key']), 'STRING')[0] === @right['key']);
});
```
into:
```java
array @arr = array(array('key': 'Hello'), array('key': 'World'), array('key': 'Aaahhh'));
@arr = array_sort(@arr, closure(@left, @right) {
    return(string_compare(@left['key'], @right['key']));
});
```

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
+++ b/src/main/java/com/laytonsmith/core/functions/ArrayHandling.java
@@ -1828,9 +1828,12 @@ public class ArrayHandling {
 						} else {
 							value = -1;
 						}
+					} else if(c.isInstanceOf(CInt.TYPE)) {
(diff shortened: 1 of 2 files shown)


Pieter12345:
As a result from a discussion with @PseudoKnight, we decided that this PR should also include a `string_compare_ic()` function that works the same as `string_compare()`, except that it ignores casing.

PR #1382, opened 2024-06-10
Add item_in_hand and hand fields to block_place event.

Lildirt:
Adds those two keys to the event array. `item_in_hand` is a full item stack, whereas `hand` is just `main_hand`/`off_hand`.


PseudoKnight:
I think 'item_in_hand' could be shortened to just 'item'. What do you think?

Other than that, this looks good to go. I checked and it should work on all supported versions of Paper/Spigot.

Pieter12345:
> I think 'item_in_hand' could be shortened to just 'item'. What do you think?
> 
> Other than that, this looks good to go. I checked and it should work on all supported versions of Paper/Spigot.

I want to add the suggestion to use `held_item`, since it clearly indicates what it represents. But that might not be consistent with most other events using `item`. You can pick any, I just wanted to add the suggestion.

Lildirt:
> I think 'item_in_hand' could be shortened to just 'item'. What do you think?
> 
> Other than that, this looks good to go. I checked and it should work on all supported versions of Paper/Spigot.

Good point. Looks like `item` (vs `held_item`) is the most consistent for events. Updated the commit to use `item` instead.

PR #1384, opened 2024-06-28
add item_pre_anvil event

Lildirt:
Kind of flying blind with creating a new event. It's been many years since I've written one, in an extension. :joy:
Let me know if there's any validation you'd like for me to do on my end.

Example payload (pretty printed):
```
event_type: item_pre_anvil
first_item: {meta: {damage: 0, display: null, enchants: {silk_touch: {elevel: 1}, unbreaking: {elevel: 3}}, flags: {}, lore: null, model: null, modifiers: null, repair: 0, tags: null, unbreakable: false}, name: DIAMOND_PICKAXE, qty: 1}
item_repair_cost: 10
level_repair_cost: 10
macrotype: inventory
max_repair_cost: 40
result: {meta: {damage: 0, display: null, enchants: {efficiency: {elevel: 4}, silk_touch: {elevel: 1}, unbreaking: {elevel: 3}}, flags: {}, lore: null, model: null, modifiers: null, repair: 1, tags: null, unbreakable: false}, name: DIAMOND_PICKAXE, qty: 1}
second_item: {meta: {damage: 0, display: null, enchants: {efficiency: {elevel: 4}, unbreaking: {elevel: 3}}, flags: {}, lore: null, model: null, modifiers: null, repair: 0, tags: null, unbreakable: false}, name: DIAMOND_PICKAXE, qty: 1}
viewers: {Lildirt}
```

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
@@ -40,7 +40,7 @@ public class BukkitMCAnvilInventory extends BukkitMCInventory implements MCAnvil
 
 	@Override
 	public int getRepairCostAmount() {
-		return ai.getRepairCost();
+		return ai.getRepairCostAmount();
 	}
 
 	@Override
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
@@ -1,0 +1,80 @@
+package com.laytonsmith.abstraction.bukkit;
+
+import com.laytonsmith.abstraction.MCAnvilInventory;
+import com.laytonsmith.abstraction.MCItemStack;
+import org.bukkit.inventory.AnvilInventory;
+
+public class BukkitMCAnvilInventory extends BukkitMCInventory implements MCAnvilInventory {
+
+	AnvilInventory ai;
+
+	public BukkitMCAnvilInventory(AnvilInventory inventory) {
+		super(inventory);
+		ai = inventory;
+	}
+
+	@Override
+	public MCItemStack getFirstItem() {
+		return new BukkitMCItemStack(ai.getItem(0));
+	}
+
+	@Override
+	public MCItemStack getSecondItem() {
+		return new BukkitMCItemStack(ai.getItem(1));
+	}
+
+	@Override
+	public MCItemStack getResult() {
+		return new BukkitMCItemStack(ai.getItem(2));
+	}
+
+	@Override
+	public int getMaximumRepairCost() {
+		return ai.getMaximumRepairCost();
+	}
+
+	@Override
+	public int getRepairCost() {
+		return ai.getRepairCost();
+	}
+
+	@Override
+	public int getRepairCostAmount() {
+		return ai.getRepairCost();
+	}
+
+	@Override
+	public String getRenameText() {
+		return ai.getRenameText();
+	}
+
+	@Override
+	public void setFirstItem(MCItemStack stack) {
+		ai.setItem(0, ((BukkitMCItemStack) stack).asItemStack());
+	}
+
+	@Override
+	public void setSecondItem(MCItemStack stack) {
+		ai.setItem(1, ((BukkitMCItemStack) stack).asItemStack());
+	}
+
+	@Override
+	public void setResult(MCItemStack stack) {
+		ai.setItem(2, ((BukkitMCItemStack) stack).asItemStack());
+	}
+
+	@Override
+	public void setMaximumRepairCost(int levels) {
+		ai.setMaximumRepairCost(levels);
+	}
+
+	@Override
+	public void setRepairCost(int levels) {
+		ai.setRepairCost(levels);
+	}
+
+	@Override
+	public void setRepairCostAmount(int levels) {
+		ai.setRepairCostAmount(levels);
+	}
+}
--- a/src/main/java/com/laytonsmith/abstraction/events/MCPrepareAnvilEvent.java
+++ b/src/main/java/com/laytonsmith/abstraction/events/MCPrepareAnvilEvent.java
@@ -1,12 +1,10 @@
 package com.laytonsmith.abstraction.events;
 
-import com.laytonsmith.abstraction.MCHumanEntity;
+import com.laytonsmith.abstraction.MCPlayer;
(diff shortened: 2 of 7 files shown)


src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	@Override
+	public int getRepairCost() {
+		return ai.getRepairCost();
+	}
+
+	@Override
+	public int getRepairCostAmount() {
+		return ai.getRepairCost();
```
Pieter12345, line 43: I believe that this should call `getRepairCostAmount()` instead.

src/main/java/com/laytonsmith/abstraction/events/MCPrepareAnvilEvent.java
```diff
@@ -0,0 +1,16 @@
+package com.laytonsmith.abstraction.events;
+
+import com.laytonsmith.abstraction.MCHumanEntity;
+import com.laytonsmith.abstraction.MCInventory;
+import com.laytonsmith.abstraction.MCItemStack;
+
+import java.util.List;
+
+public interface MCPrepareAnvilEvent extends MCInventoryEvent {
+	@Override
```
Pieter12345, line 10: What is this override doing here?

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
```diff
@@ -987,4 +989,90 @@ public Version since() {
...
+		@Override
+		public String getName() {
+			return "item_pre_anvil";
+		}
+
+		@Override
+		public String docs() {
+			return "{}"
+					+ " Fires when a recipe is formed in an anvil, but the result has not yet been clicked."
+					+ " { viewers: all human entities looking at this anvil "
```
Pieter12345, line 1005: I would word this as `all players viewing this anvil's interface` or something similar that is a bit more explicit.

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
```diff
@@ -987,4 +989,90 @@ public Version since() {
...
+					+ " | first_item: the first item in the anvil"
+					+ "	| second_item: the second item in the anvil"
+					+ " | result: the result of the anvil"
```
Pieter12345, line 1008: I would word this as:
`the first ingredient in the anvil recipy`
`the second ingredient in the anvil recipy`
`the result of the anvil recipy`
Or something alike. Again a bit more explicit, leaving less room for misinterpretation (what is the result of an anvil?).

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
```diff
@@ -987,4 +989,90 @@ public Version since() {
...
+		}
+
+		@Override
+		public BindableEvent convert(CArray manualObject, Target t) {
+			throw ConfigRuntimeException.CreateUncatchableException("Unsupported operation.", Target.UNKNOWN);
+		}
+
+		@Override
+		public Map<String, Mixed> evaluate(BindableEvent event) throws EventException {
+			if(event instanceof MCPrepareAnvilEvent) {
```
Pieter12345, line 1031: Java now has a nice language feature that you can use here.
```java
if(event instanceof MCPrepareAnvilEvent) {
    MCPrepareAnvilEvent e = (MCPrepareAnvilEvent) event;
```
is equivalent to:
```java
if(event instanceof MCPrepareAnvilEvent e) {
```

It's up to you to decide whether or not you want to use that, but it's been used more and more in CH.

Pieter12345 (commented):
Quick review without having compiled or executed the code. It looks like you were successful at sticking to the project structure! Thanks for contributing :)

src/main/java/com/laytonsmith/abstraction/bukkit/BukkitMCAnvilInventory.java
Lildirt: Indeed it should! Updated.

src/main/java/com/laytonsmith/abstraction/events/MCPrepareAnvilEvent.java
Lildirt: Uh, nothing. Actually, that entire `getInventory()` definition doesn't need to exist. It's defined in `MCInventoryEvent`. Removed it.

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
Lildirt: Verbage updated.

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
Lildirt: Verbage updated.

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
Lildirt: Sure! I've updated it to use the short-hand.

PseudoKnight:
Looks like you matched 'viewers' from item_pre_craft, whereas we should probably get a player for both of these from the InventoryView. I don't think there can be multiple viewers, at least for these events on these GUIs. I would recommend removing 'viewers' and add 'player' for this event.

There's one hitch. InventoryView was changed from an abstract class to an interface a couple of weeks ago. Directly using methods from that interface may not work on older versions of the server. Might need reflection. Something like:

```
	@Override
	public MCHumanEntity getPlayer() {
		return new BukkitMCHumanEntity(ReflectionUtils.invokeMethod(e.getView(), "getPlayer"));
	}
```

PseudoKnight:
Let me test the other event to see if reflection is needed here too.

PseudoKnight:
Actually can probably just do e.getViewers.get(0) instead of reflection. Or should we have that as a fallback for getPlayer() on older versions?

Lildirt:
> Actually can probably just do e.getViewers.get(0) instead of reflection. That's probably the way to go.

Pushed a commit. Something like that?

PseudoKnight:
That looks fine. It looks impossible to have more than one viewer for an anvil. Unlike shared inventory views, the list of tracked viewers of an anvil is created every time a call to open an anvil is used.

PseudoKnight:
Just noticed you have question marks in the description of the repair costs. Perhaps they were temporary?

PR #1385, opened 2024-07-04
add modify for item_pre_anvil's item_repair_cost/level_repair_cost

Lildirt:
While writing some code around `item_pre_anvil`, I noticed you can't actually *click* the result slot if the result tool has no enchantments on it. It becomes a nothing click if the level cost isn't set.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
@@ -1059,6 +1059,17 @@ public class InventoryEvents {
 					e.setResult(ObjectGenerator.GetGenerator().item(value, t));
 					return true;
 				}
+
+				MCAnvilInventory anvil = (MCAnvilInventory) e.getInventory();
+				if(key.equalsIgnoreCase("level_repair_cost")) {
+					anvil.setRepairCost(ArgumentValidation.getInt32(value, t));
+					return true;
+				}
+
+				if(key.equalsIgnoreCase("item_repair_cost")) {
+					anvil.setRepairCostAmount(ArgumentValidation.getInt32(value, t));
+					return true;
+				}
 			}
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


PseudoKnight:
Oh good. This is useful for other cases too. Add these to the mutable fields in documentation.

PseudoKnight:
Does setMaximumRepairCost() work here too? If so, that would be worth adding as well if people want to uncap repair costs.

PseudoKnight:
Just noticed repair get|setRepairCostAmount() was added in 1.18.1. We'll need to fix this for earlier versions. If you want to do it, you just have to check `Static.getServer().getMinecraftVersion().gte(MCVersion.MC1_18_1)` before using those methods in evaluate() and modifyEvent(). Then I usually add (MC 1.18.1+) to the data field in the docs.

Lildirt:
> Does setMaximumRepairCost() work here too? If so, that would be worth adding as well if people want to uncap repair costs.

Sort of. It does work, and the max cost is respected on the server side. I can smith a tool with a level cost of 50 and a max cost of 55, but the client side will always show "Too Expensive!" for anything that breaks the Vanilla limit.

I can still include it. I don't think there's any way to fix the "Too Expensive!" bit on the client-side though.

[Screenshot](https://dirt.wtf/7T2QcjPfytxNk1Ox.png)

Lildirt:
Pushed the suggested changes. Included a disclaimer on the docs related to my last comment.

PseudoKnight:
That's a shame. So not only is the cost higher, but it's a mystery! One could change that text with a resource pack, probably, but it would be impossible to show the actual cost in the GUI. Would have to print it to chat or something. Still, probably worth including.

Anyway, looks good!

PseudoKnight:
I guess the one thing a user could do is cap the repair cost to the lowest visible value. Not sure if that's 39 or 40. But if they want to make a player pay for the extra repair, they'll have to settle for some other visual indicator or the mystery.

PR #1386, opened 2024-07-06
add item_pre_grindstone, item_pre_smithing; partial fix modify_event for item_pre_craft

Lildirt:
- Adding `modify_event('result', ...);` for `item_pre_craft`.
- Adding `item_pre_grindstone` (`PrepareGrindstoneEvent`) event.
- Adding `item_pre_smithing` (`PrepareSmithingEvent`) event.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
@@ -943,7 +943,6 @@ public class InventoryEvents {
 				for(int i = 0; i < mi.length; i++) {
 					matrix.set(i, ObjectGenerator.GetGenerator().item(mi[i], t), t);
 				}
-				ret.put("inventorytype", new CString(e.getInventory().getType().toString(), t));
 				ret.put("matrix", matrix);
 				ret.put("result", ObjectGenerator.GetGenerator().item(e.getInventory().getResult(), t));
 				return ret;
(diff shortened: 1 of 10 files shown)


src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
```diff
@@ -938,6 +943,7 @@ public Map<String, Mixed> evaluate(BindableEvent event) throws EventException {
 				for(int i = 0; i < mi.length; i++) {
 					matrix.set(i, ObjectGenerator.GetGenerator().item(mi[i], t), t);
 				}
+				ret.put("inventorytype", new CString(e.getInventory().getType().toString(), t));
```
PseudoKnight, line 946: This isn't documented. However, is this even needed for something? It only returns CRAFTING or WORKBENCH right now, which can already be inferred by the matrix size, which is the only meaningful difference as far as this event is concerned. Maybe if this fired for CRAFTER too, but it doesn't. 

In any case, one of the other needs to change: document this or remove it.

Lildirt: Removed.

src/main/java/com/laytonsmith/core/events/drivers/InventoryEvents.java
```diff
@@ -1068,4 +1053,159 @@ public Version since() {
...
+				MCSmithingInventory smithing = (MCSmithingInventory) e.getInventory();
+				Map<String, Mixed> ret = evaluate_helper(e);
+
+				ret.put("player", new CString(e.getPlayer().getName(), Target.UNKNOWN));
+				ret.put("first_item", ObjectGenerator.GetGenerator().item(smithing.getInputTemplate(), Target.UNKNOWN));
+				ret.put("second_item", ObjectGenerator.GetGenerator().item(smithing.getInputEquipment(), Target.UNKNOWN));
+				ret.put("third_item", ObjectGenerator.GetGenerator().item(smithing.getInputMaterial(), Target.UNKNOWN));
+				try {
+					ret.put("recipe", ObjectGenerator.GetGenerator().recipe(smithing.getRecipe(), Target.UNKNOWN));
+				} catch(NullPointerException ex) {
```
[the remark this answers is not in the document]
Lildirt: Yup, your assumption is correct. I've moved the null check to `BukkitMCSmithingInventory` as suggested.

PR #1387, opened 2024-07-17
Add player_bucket_fill and player_bucket_empty events.

Lildirt:
Payloads:
```
event_type: player_bucket_empty
hand: main_hand
item: {meta: null, name: BUCKET, qty: 1}
location: {0: -295.0, 1: 118.0, 2: -29.0, 3: world, world: world, x: -295.0, y: 118.0, z: -29.0}
macrotype: player
player: Lildirt
```

```
event_type: player_bucket_fill
hand: main_hand
item: {meta: null, name: WATER_BUCKET, qty: 1}
location: {0: -295.0, 1: 118.0, 2: -29.0, 3: world, world: world, x: -295.0, y: 118.0, z: -29.0}
macrotype: player
player: Lildirt
```

I debated on whether to include `item` with this, as it is the result item *after* the event has processed. However, getting the held item of the player will give you what they were holding *before* the event happened, so I suspect it has some use.

Also, I mocked the events around pistons for this kind of format, which is why I have a base event that is being extended by both.


PseudoKnight:
Looks good to me. It's well thought out. I agree with your assessment of 'item'. I guess the key should be changed to be more clear, but it's not that important. I also agree that there doesn't seem to be a point to including the clicked block and face. We can always add it later if it proves to be useful in some way, but player_interact can handle that already.

Anyway, I'm ready to pull this.

PseudoKnight:
Maybe some small tweaks to the docs.

I would use "into the world" for the empty event description.

Even though it matches other events, could explicitly say that 'hand' is either 'main_hand' or 'off_hand'.

PseudoKnight:
Oh, getHand() was only added to these events in 1.19.2.

Lildirt:
Added 1.19.2+ gatekeep for `getHand()` and your suggested doc change.

PR #1388, opened 2024-07-26
Fix location shift in block_break so dropped items aren't placed into adjacent blocks.

Lildirt:
If you call `modify_event('drops', <any item array>)` for `block_break`, there's a bit of a 'hack' in place to emulate this since [`BlockBreakEvent`](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/block/BlockBreakEvent.html) doesn't support rewriting drops natively.

Since we offset the location then do a `dropItemNaturally` call, it can sometimes send the item *into* another block, which causes it to shoot to the surface if they're underground. Not ideal.

I've attached links to two gifs. They're a bit large (and might play slow until they download fully). Hopefully it visualizes what happens.

Before: (~13MB) https://dirt.wtf/1ebnnKc8Pne20dm0.gif

After: (~8MB) https://dirt.wtf/VFB66bDgc6zGsRFa.gif

### DIFF
--- a/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitBlockListener.java
+++ b/src/main/java/com/laytonsmith/abstraction/bukkit/events/drivers/BukkitBlockListener.java
@@ -86,7 +86,6 @@ public class BukkitBlockListener implements Listener {
 				Block block = chevent.getBlock();
 				block.setType(Material.AIR);
 				Location loc = block.getLocation();
-				loc.add(0.5, 0.5, 0.5);
 				if(chevent.isDropItems()) {
 					for(MCItemStack item : bbe.getDrops()) {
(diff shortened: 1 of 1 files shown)


PseudoKnight:
This was probably due to a misunderstanding of what dropItemNaturally() does at the time of writing this. Even though it's not documented, it seems to assume the Location is a Block location, so it adds 0.5 (with the random offset) to the coordinates.

It appears Minecraft vanilla also subtracts half the item entity height (i.e. 0.125) from the randomized y position, which dropItemNaturally() does not do. It's not a big deal, but we could fix that too.

Lildirt:
:+1: Subtracting 0.125 from Y.

PseudoKnight:
Want to rewrite the description of this issue before I pull? I can do it if you want. Something along the lines of fixing the position of modified block drops in block_break.

PR #1398, opened 2025-04-25
Sleeping ignored

malonnnn:
This adds getter/setter for setSleepingIgnored(). I am a barely functioning troglodyte. I have no idea if I did this correctly, but it's working for me, so I hope that's something


malonnnn:
Does one of those errors mean I built it with the wrong sdk?

LadyCailin:
> Does one of those errors mean I built it with the wrong sdk?

No. I only gave your PR a very quick glance, but that error means the pipeline is screwed up. Let me take a look.

PR #1399, opened 2025-06-11
Add vscode presetting and entity toggle swim event

Anatoliy057:
I added presets for vscode, specifically the code formatting configuration. Tried to make it the best fit for the project, but there are inconsistencies in it project itself, and some of the rules just can't be turned off. Anyway this is only for vscode and only if you execute formating manually.

### DIFF
--- a/src/main/java/com/laytonsmith/core/events/drivers/EntityEvents.java
+++ b/src/main/java/com/laytonsmith/core/events/drivers/EntityEvents.java
@@ -39,6 +39,7 @@ import com.laytonsmith.abstraction.events.MCEntityPortalEvent;
 import com.laytonsmith.abstraction.events.MCEntityRegainHealthEvent;
 import com.laytonsmith.abstraction.events.MCEntityTargetEvent;
 import com.laytonsmith.abstraction.events.MCEntityToggleGlideEvent;
+import com.laytonsmith.abstraction.events.MCEntityToggleSwimEvent;
 import com.laytonsmith.abstraction.events.MCEntityUnleashEvent;
 import com.laytonsmith.abstraction.events.MCEntityPotionEffectEvent;
 import com.laytonsmith.abstraction.events.MCFireworkExplodeEvent;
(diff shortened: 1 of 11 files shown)


Anatoliy057:
@Pieter12345 these files prompt vscode for a list of recommended extensions, one of which you mentioned and the settings for them (literally a couple of things and file paths).

the only thing questionable is the config file formatting, but added just as a why not

src/main/java/com/laytonsmith/core/events/drivers/EntityEvents.java
```diff
@@ -2053,6 +2055,91 @@ public Driver driver() {
...
+					+ " | player: <macro> The player triggering the event}"
+					+ " This event is called when an entity's swimming status is toggled"
+					+ " {id: The entityID of the entity | type: The entity type of the entity |"
+					+ " swimming: true if the entity is entering swimming mode, false if the entity is leaving it |"
+					+ " player: If the entity is a player, this will contain their name, otherwise null}" + " {}"
+					+ " {}";
+		}
+
+		@Override
+		public boolean matches(Map<String, Mixed> filter, MCEntityToggleSwimEvent e) throws PrefilterNonMatchException {
```
LadyCailin, line 2078: Instead of matches(), override the getPrefilterBuilder() method. This new method is much faster at runtime, and offers additional benefits such as better documentation. There are several examples of this, just search for "getPrefilterBuilder" to see some examples you can learn from (player_chat for instance has been converted), or see the PrefilterBuilder class. Once you implement that method, the matches() method goes away.

LadyCailin (changes requested):
Everything looks good except the matches() method. Let me know if you have questions on how to implement that.

src/main/java/com/laytonsmith/core/events/drivers/EntityEvents.java
Anatoliy057: Changed it, but for a player that may not exist and the id of the entity didn't appear to have any examples (or I didn't notice). For the new filter I used the documentation from a similar one, as I don't quite understand how their tags work.

LadyCailin (approved):

PR #1420, opened 2026-07-17
Typecheck procedure signatures + Various changes

Pieter12345:
# Changes
- Typecheck procedure signatures. This assumes that procedure signatures cannot be redefined after the initial declaration, which does not hold for the current implementation. Enforcing this should be done at a later point as this potentially breaks user scripts.
- Disallow mandatory-after-optional params in proc signature typechecking.
- Prevent injected `@arguments` variable overwriting an explicitly declared procedure parameter named `@arguments`.
- Initialize procedure varargs parameter with an empty array if it matches no arguments. Prevents varargs parameter value being `CNull.UNDEFINED` when it has no matching arguments.
- Add compile error on missing returns in procedures.
- Implement function signatures and `getReturnType()` in control flow manipulating functions.
- Throw exception when invoking proc forward declaration (opposed to not doing anything and returning void, regardless of the declared proc return type).
- Add none type and missing return false error tests. These ensure that mentioned typechecking and missing return error changes do not break the obvious use-cases that can be expected to be affected.
- Implement function signature none type allowed boolean per Param instead of per FunctionSignature. This is reflected in the function signatures string.
- Apply none type allowed boolean in FunctionSignature to all type of Param instead of only to varparams.
- Update function signatures of `dand()` and `dor()`. Fixes NPE when calling `dor()` without arguments (given that SA is enabled).

# WIP
This PR is marked as work-in-progress for the following reasons that should be addressed before merging:
- ~`dor()` and `dand()` typechecking implementation needs to be implemented in order to prevent missing return errors for returns within them that are guaranteed to run.~
- ~This PR should be tested against large code bases to identify potentially undesired breaking changes.~

### DIFF
--- a/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
+++ b/src/main/java/com/laytonsmith/core/functions/BasicLogic.java
@@ -16,7 +16,6 @@ import com.laytonsmith.core.compiler.analysis.Scope;
 import com.laytonsmith.core.compiler.analysis.StaticAnalysis;
 import com.laytonsmith.core.compiler.signature.FunctionSignatures;
 import com.laytonsmith.core.compiler.signature.SignatureBuilder;
-import com.laytonsmith.core.constructs.Auto;
 import com.laytonsmith.core.constructs.CArray;
 import com.laytonsmith.core.constructs.CBoolean;
 import com.laytonsmith.core.constructs.CClassType;
@@ -28,6 +27,7 @@ import com.laytonsmith.core.constructs.CNumber;
 import com.laytonsmith.core.constructs.CString;
 import com.laytonsmith.core.constructs.CSymbol;
 import com.laytonsmith.core.constructs.CVoid;
+import com.laytonsmith.core.constructs.InstanceofUtil;
 import com.laytonsmith.core.constructs.Target;
 import com.laytonsmith.core.environments.Environment;
 import com.laytonsmith.core.environments.GlobalEnv;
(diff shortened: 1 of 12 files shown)


src/main/java/com/laytonsmith/core/functions/BasicLogic.java
```diff
@@ -1570,8 +1600,35 @@ public FunctionSignatures getSignatures() {
...
-					.setNoneIsAllowed(true).build();
+			return new SignatureBuilder(CClassType.AUTO)
+					.param(Mixed.TYPE, "val", "The first value.")
+					.varParam(Mixed.TYPE, "vals", "The values.")
+					.param(Mixed.TYPE, "termVal", "The final terminating (non-returning) value.", true, true).build();
+		}
+
+		@Override
+		public CClassType getReturnType(Target t, List<CClassType> argTypes,
+				List<Target> argTargets, Environment env, Set<ConfigCompileException> exceptions) {
```
LadyCailin, line 1596: This can be consolidated into a helper method that both dand and dor call. Actually, this can be called (perhaps with a parameter) in ifelse and switch too. Especially since it has a TODO, best to consolidate it into a separate method.

Pieter12345: There are several places where either the first type in a list of the most specific type is returned as temporary measure for missing multi-types. I've created static helper methods in `CClassType` with documentation to keep these in a single place. That reduces the `getReturnType()` implementation to two lines of code, for which I don't think another helper method is useful. These `getReturnType()` implementations are expected to be removed as soon as 'fancy' function signatures with generics are implemented.

